PR #4, opened 2016-10-29
Refactoring

ellmetha:
Note that this will ease the support of multiple containers per project.


hsoft:
Is there a specific reason for wanting to support Python 2? I understand that, at this stage, the cost for supportig it is low, but as we go futher, it might require more effort.

I have trouble imagining a case where the lack of python 3 support would be detrimental to adoption.

nomad/constants.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+
+
+# CONTAINERS
+# --
+
+CONTAINER_STOPPED = 102
```
hsoft: You didn't like the pseudo-enum style?

nomad/container.py
```diff
@@ -0,0 +1,250 @@
...
+        if barebone is None:  # None == only if the container isn't provisioned.
+            barebone = not self.is_provisioned
+
+        if barebone:
+            logger.info('Doing bare bone setup on the machine...')
+            prepare_debian(self._container)
+
+        logger.info('Provisioning container "{name}"...'.format(name=self.name))
+        for provisioning_item in self.options.get('provisioning', []):
+            logger.info('Provisioning with {0}'.format(provisioning_item['type']))
```
hsoft: Not that it matters much, but I'm curious as to why you've added the explicit `0` in `{0}`.

nomad/constants.py
ellmetha: Well depending on the Python 3 versions we want to support, we could have used the built-in ``Enum``. ;-) But I don't think it is necessary to wrap this kind of values in classes for now. We could also put these values in the new ``Container`` class. What do you think?

nomad/container.py
ellmetha: Just an old habit. ;-)

nomad/provision.py
```diff
@@ -35,14 +42,13 @@ def set_static_ip_on_debian(container, ip, gateway):
...
 
+
 def provision_with_ansible(container, provisioning_item):
     assert provisioning_item['type'] == 'ansible'
     ip = get_ipv4_ip(container)
     with tempfile.NamedTemporaryFile() as tmpinv:
         tmpinv.write("{} ansible_user=root".format(ip).encode('ascii'))
         tmpinv.flush()
         cmd = "ANSIBLE_HOST_KEY_CHECKING=False ansible-playbook -i {} {}".format(tmpinv.name, provisioning_item['playbook'])
-        print(cmd)
```
ellmetha: You're right. I'll add a line to log this value.

ellmetha:
> Is there a specific reason for wanting to support Python 2? I understand that, at this stage, the cost for supportig it is low, but as we go futher, it might require more effort.

It could be useful for systems where Python 3 is not the default, preferred version. ;-) I think that if we start supporting both versions of Python in early stages, it will be easier to maintain in the long run.

hsoft:
Considering that this is a development tool, I'm not sure that it's worth it. After all, Ubuntu 16.04 switched to a python 3 default. Sure, supporting python2 isn't a big deal, but it's the *little things* that end up piling up. One of them being that we can't use `pathlib`.

We could say "oh, let's wait until it's actually complicated to strip python 2 support", but somehow, I suspect that it will be harder to make that decision later because we'll wonder if we break someone's workflow by doing so.

ellmetha:
OK. ;-) I see your point. Let's remove py2/py3 compatible code.

hsoft:
Works fine, looks fine.

PR #6, opened 2016-11-07
Fix a KeyError occuring when the image is not available...

ellmetha:
... or if there is an error in the container's dict config


hsoft:
I don't think we need PRs for this kind of stuff.

PR #7, opened 2016-11-07
Add pull mode

ellmetha:
Add support for pull mode: this allows to create containers without having to copy them. The image are fetched from the linuxcontainers image server by default (but this can be configured of course).

One thing: for now the "local" mode remains the default mode for LXD-Nomad, but I think that the "pull" mode should be the default mode because it allows a workflow that is much more simple than the "local" mode (you don't have to initially copy the image and its aliases before using the nomad command).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,11 @@ provisioning:
 *something* out of `nomad` commands made in the same folder.
 
 You can also choose to define a container that will be created by pulling directly one of the image
-hosted on https://images.linuxcontainers.org/: this is the "pull" mode:
+hosted on https://images.linuxcontainers.org/. This is the "pull" mode:
 
 ```
 name: myproject
-image: ubuntu/xenial/amd64
+image: ubuntu/xenial
 mode: pull
 hostnames:
   - myproject.local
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,6 +38,29 @@ provisioning:
 ... and that you have a pre-configured `jessie` container that works, you should be able to get
 *something* out of `nomad` commands made in the same folder.
 
+You can also choose to define a container that will be created by pulling directly one of the image
+hosted on https://images.linuxcontainers.org/: this is the "pull" mode:
+
+```
+name: myproject
+image: ubuntu/xenial/amd64
```
ellmetha: Yep. I totally agree. It turns out that I had a problem using both pull mode and "implicit" aliases like "ubuntu/xenial" (the container was not created and and error occurred). Using the "simplestreams" protocol instead of "lxd" seems to fix the problem.

hsoft: I'm trying it right now and experience the same issue. What about making `simplestreams` the default protocol? According to my tests here, that's what `lxc` seems to use anyway. If I do `lxc image copy images:debian/stretch local: --copy-aliases`, `lxc image info debian/stretch` tells me that its protocol is `simplestreams`. We might as well mimick `lxc`...

ellmetha: I agree. I'll do that.

PR #9, opened 2016-12-13
Add support for multiple containers

ellmetha:


nomad/conf/config.py
```diff
@@ -55,10 +58,30 @@ def load(self):
...
+        The dictionary that is returned contains the container's configuration and the global config
+        values. These global values can be defined outside of the scope of the container's config
+        and can be used by each container.
+        """
+        container_config = {}
+        container_config.update(container_dict)
+        global_config = dict(self._dict)
+        del global_config['containers']
+        for k, v in global_config.items():
+            if k not in container_config:
```
hsoft: There's `setdefault()` for that pattern.

nomad/conf/config.py
```diff
@@ -55,10 +58,30 @@ def load(self):
...
+                container_config[k] = v
+        return container_config
```
hsoft: ... in fact, why not simply:

```
result = dict(self._dict)
del result['container']
result.update(container_dict)
return result
```

???

ellmetha: Because `update` will overwrite the config values that could've been overridden at the container level. :wink:

PR #12, opened 2017-01-02
Improve container names for multi-containers

ellmetha:


nomad/constants.py
```diff
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
 
 
+# PROJECTS
+# --
+
+METADATA_BASE_DIR = '.nomad'
```
hsoft: I really liked the idea of leveraging LXD's metadata system to avoid polluting project folders. I'm trying to see, in this changeset, what exactly justifies the creation of this folder, but I can't see it. Can you help me?

nomad/container.py
```diff
@@ -165,18 +173,26 @@ def is_stopped(self):
...
+        This name id supposed to be unique among all the containers managed by LXD.
+        """
+        # Note: all container names must be a valid hostname! That is: maximum 63 characters, no
+        # dots, no digit at first position, made entirely of letters/digits/hyphens, ...
+        if not hasattr(self, '_lxd_name'):
+            lxd_name_prefix = '{project_slug}-{name}'.format(
+                project_slug=slugify(self.project_name), name=self.name)
+            container_id = self.id
+            self._lxd_name = '{prefix}-{id}'.format(
+                prefix=lxd_name_prefix[:63 - len(container_id)], id=container_id)
```
hsoft: How about only appending that ID when there's actually a name clash? What we're doing here is to burden our name with a hash for a situation that will rarely occur.

nomad/container.py
```diff
@@ -188,34 +204,50 @@ def _assign_free_static_ip(self):
...
+        The <key, value> pair is created if it cannot be retrieved from the metadata file.
+        """
+        metadata_filename = self.name + '.json'
+        metadata_filepath = os.path.join(
+            self.homedir, constants.METADATA_CONTAINERS_DIR, metadata_filename)
+        metadata_exists = os.path.exists(metadata_filepath)
+
+        metadata = {}
+        if metadata_exists:
+            with open(metadata_filepath, 'r+') as fp:
```
hsoft: Why the `+`?

nomad/container.py
```diff
@@ -188,34 +204,50 @@ def _assign_free_static_ip(self):
...
+                metadata = json.loads(fp.read())
+
+        value = metadata.get(key, new_value)
+        metadata[key] = value
+
+        with open(metadata_filepath, 'w+') as fp:
```
hsoft: Again, why the `+`?

nomad/utils/text.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+_punct_re = re.compile(r'[\t !"#$%&\'()*\-/<=>?@\[\\\]^_`{|},.]+')
+
+
+def slugify(text):
+    """ Returns a ASCII-only slug version of the considered text. """
+    result = []
+    for word in _punct_re.split(text.lower()):
+        word = unicodedata.normalize('NFKD', word).encode('ascii', 'ignore').decode('ascii')
```
hsoft: I'm uneasy about this `ignore` flag. Since the project name's only purpose is to help determine a container name, what about enforcing (and documenting) a 7bit ASCII range? I don't see any reason to go beyond that range and I feel like this may cause some unforeseen problems.

hsoft: Besides, it's the same thing with `_punct_re`. Why not simply restrict what we can use in our names? To quote Bill Gates, something like `\w_-` ought to be enough for anybody.

nomad/constants.py
ellmetha: The goal here is to store a unique ID associated with each container managed by nomad.

Currently we are using a custom identifier that is associated with our container using the LXD's metadata system (see [container.py#L241](https://github.com/lxd-nomad/lxd-nomad/blob/master/nomad/container.py#L241)). The name of each container is generated at creation time by checking if there clashes with existing containers, because we want our container names to be unique (see [container.py#L205](https://github.com/lxd-nomad/lxd-nomad/blob/master/nomad/container.py#L205)). The problem with this approach is that it forces us to things like this to retrieve a container that is already created (see [container.py#L193](https://github.com/lxd-nomad/lxd-nomad/blob/master/nomad/container.py#L193)): 

```python
container = None
for _container in self.client.containers.all():
    if _container.config.get('user.nomad.nid') == str(self.nid):
        container = _container
```

This is not the good approach because we can't really get a specific container managed by nomad without checking the metadata of all the available containers. We should use the `get` method provided by pyLXD in order to retrieve a container by its name (see https://github.com/lxc/pylxd/blob/master/doc/source/containers.rst#examples), which I think is the preferred way of doing this. But we can't do this currently because we don't know what is the name of the container we are working on (at the LXD level). We need to know what are the names of our containers in order to properly retrieve them.

It is true that we can use the LXD's metadata system to store config values associated with each containers. But we need to store the identifier that allows us to get our container by its name outside this system (otherwise we cannot retrieve our container directly with the `get` method mentioned proviously). This is why this "ID" has to be stored outside the LXD's metadata system, for example in a hidden metadata file at the root of the project's folder.

nomad/container.py
ellmetha: That's true, but the same could be said for Vagrant (which creates machine names using a similar pattern). My point is that it doesn't hurt to associate a unique identifier to every container that is created using lxd-nomad and to systematically use it to ensure that its LXD name is unique. Thus we can directly retrieve the containers associated with a specific project. The name are a bit longer, but the users should not use the "lxc" commands so often (they will always use the "local" names of the containers when running "nomad up/halt/...").

nomad/utils/text.py
ellmetha: Currently the project's name can be either manually set in the nomad file or dynamically set using the folder name of the project. eg. if the nomad.yml file is stored in "/path/to/mon projet", "mon projet" will be the name of the project (if the project name is not in the nomad.yml file of course). So we have to enforce that our LXD name contains only character that can be part of a hostname (actually all LXD names must be valid hostnames).

I agree that we should enforce some rules regarding project names and container names (when they are manually set). But this will require some validation logic in order to validate the content of a nomad file prior any operations with containers (my plan was to add validation rules for nomad files with another pull request).

But when the project's name is computed from the project's path, we should slugify this value when using it to create LXD names. :wink:

nomad/container.py
ellmetha: yup, my mistake!

nomad/container.py
ellmetha: A rest of an old implementation! My bad!

nomad/constants.py
hsoft: If I understand correctly then, this particular change (the addition of the `.nomad` folder) is not strictly required by this PR. It just happens to be bundled with it. We could have introduced this change earlier with the same justification, and we could introduce it later with the same justification.

As for the justification itself, it seems that we're faced with two "evils":

1. Enumerating through all containers at each `nomad` runs.
2. Polluting project folders (and force users to add yet another entry to all their `.gitignore`).

Unless there's a particular performance problem that hit us after a certain number of containers, I don't see how evil 1 is worse than evil 2.

There's an additional set of evils:

1. Lose container references when we move the project folder.
2. Lose container references when we do `git clean -fxd`.

But then again, it's hard to see which one is worse.

ellmetha: Hmm, perhaps we could use another strategy to ensure that our LXD name is unique (and well known by lxd-nomad).

Instead of creating a name using the patterm "{project_slug}-{container_name}-{id}" with a generated ID (which forces us to store it inside a "metadata" hidden file), we could use the same pattern with an ID corresponding to the inode number of the folder corresponding to the "homedir" of the project, thus we could be sure that the LXD name is unique.

ellmetha: @hsoft OK, I've made some changes regarding this issue. :wink:

hsoft: Interesting!

hsoft: And with the help of `user.nomad.made`, we could easily create a command to purge "orphan" containers because it's still possible to reverse-find (or *not* find) a `nomad.yml` file from a LXD container.

nomad/utils/text.py
hsoft: Now that we use inodes, isn't this argument obsolete? If we validate project/container names, we shouldn't need `slugify()` any more.

ellmetha: We are still using project names to generate our LXD hostnames (because it's a good practice to prefix our container names with the name of the related project). If the project name is set using the homedir name, this values must be "slugified". The question is: should we enforce the presence of the project's name in the config file (in that case we would assume that the related value is already a slug, so the slugify tool would be useless)?

hsoft: Oh, yes, I see, you mean https://github.com/lxd-nomad/lxd-nomad/blob/improve_multi_containers_names/nomad/cli/project.py#L20

Yes, let's enforce a project name. Using the folder name is going to give us problems. What are we going to do when chinese people run nomad in a folder with 100% chinese characters?

ellmetha: Fair enough. :wink: I'll remove the `slugify` helper, but the proper validations ruls will be added in a subsequent PR.

nomad/container.py
```diff
@@ -165,18 +168,28 @@ def is_stopped(self):
...
     @property
+    def lxd_name(self):
+        """ Returns the name of the container that is used in the scope of LXD.
+
+        This name id supposed to be unique among all the containers managed by LXD.
+        """
+        # Note: all container names must be a valid hostname! That is: maximum 63 characters, no
+        # dots, no digit at first position, made entirely of letters/digits/hyphens, ...
+        if not hasattr(self, '_lxd_name'):
+            lxd_name_prefix = '{project_slug}-{name}'.format(
```
hsoft: In single-container projects (the vast majority of use cases), we end up with duplicated project names (example: `foo-foo-20504857943`) in the LXD container names because of the way the config is passed around at instantiation.

Wouldn't it be nice if we could tidy this up? Suggestion:

1. Get rid of the creation counter logic, make `name` a required argument.
2. Assign auto-generated based on `enumerate()` at https://github.com/lxd-nomad/lxd-nomad/blob/improve_multi_containers_names/nomad/conf/config.py#L61
3. At https://github.com/lxd-nomad/lxd-nomad/blob/improve_multi_containers_names/nomad/conf/config.py#L66 override name so it's empty.
4. In this very line, don't include `name` if it's empty
5. Use `lxd_name` everywhere in our log messages.

nomad/container.py
```diff
@@ -165,18 +167,28 @@ def is_stopped(self):
...
     @property
+    def lxd_name(self):
+        """ Returns the name of the container that is used in the scope of LXD.
+
+        This name id supposed to be unique among all the containers managed by LXD.
+        """
+        # Note: all container names must be a valid hostname! That is: maximum 63 characters, no
+        # dots, no digit at first position, made entirely of letters/digits/hyphens, ...
+        if not hasattr(self, '_lxd_name'):
+            lxd_name_prefix = '{project_name}-{name}'.format(
```
hsoft: In single-container projects (the vast majority of use cases), we end up with duplicated project names (example: `foo-foo-20504857943`) in the LXD container names because of the way the config is passed around at instantiation.

Wouldn't it be nice if we could tidy this up? Suggestion:

1. Get rid of the creation counter logic, make `name` a required argument.
2. Assign auto-generated based on `enumerate()` at https://github.com/lxd-nomad/lxd-nomad/blob/improve_multi_containers_names/nomad/conf/config.py#L61
3. At https://github.com/lxd-nomad/lxd-nomad/blob/improve_multi_containers_names/nomad/conf/config.py#L66 override name so it's empty.
4. In this very line, don't include `name` if it's empty
5. Use `lxd_name` everywhere in our log messages.

ellmetha: I'll come back with a new commit but I don't see why we should use `lxd_name` in log messages (5) - we are using "local names" with the nomad command (which seems fine IMO).

My proposal: we could assign an auto-generated name if needed as you said (2) - eg. "default" - if we are considering a single-container project. If the project contains more than one container, we could enforce the definition of container names in the configuration file... or we could just enforce container names in all cases.

hsoft: oh, yes, I was forgetting about referencing containers in the CLI... You are right. If we make project name mandatory, we might as well do the same for containers. That would leave us with only (3) and (4) to do.

ellmetha: OK. I've added a commit to enforce container names. (3) and (4) are no longer required since the name of an implicit container (that is, a container which is not defined under the "containers" section) will automatically default to "default". I think this is the correct behavior for now.

Regarding the validation rules problem and how to "properly" enforce project names / container names (without presenting a KeyError exception to the user)... these aspects should be improved in a subsequent PR! :wink:

hsoft:
Good, stuff, thanks!

PR #13, opened 2017-01-06
Add schema validation for nomad files

ellmetha:


nomad/conf/validators.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
hsoft: I afford comments on minor stuff because I don't have much to comment on the rest. Why do you keep specifying source encoding in your units? it's utf-8 by default in python 3.

nomad/conf/validators.py
```diff
@@ -0,0 +1,31 @@
...
+
+import re
+
+from voluptuous.schema_builder import message
+from voluptuous.validators import truth
+
+
+hostname_part_re = re.compile(r'(?!-)[A-Z\d-]{1,63}(?<!-)$', re.IGNORECASE)
+lxd_identifier_re = re.compile(r'^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$', re.IGNORECASE)  # noqa
```
hsoft: Since you have that `IGNORECASE` flag, can't we do like in `hostname_part_re` and use `A-Z` only?

nomad/conf/validators.py
ellmetha: You're right. This is just that old habit of writing Python 2/3 compatible code.

hsoft:
Also: this looks like a nice lib, thanks for the discovery!

nomad/conf/validators.py
ellmetha: @hsoft I've fixed that. :wink:

PR #15, opened 2017-01-17
Added some colors for warnings and errors

ellmetha:


hsoft:
Looks good to me, but now it's your turn to review my fix.

hsoft:
Oops, my fix was not well tested, it doesn't work. fixing...

hsoft:
Now, that's better

nomad/project.py
```diff
@@ -61,7 +61,8 @@ def shell(self, container_name=None):
             raise ProjectError(
                 'This action requires a container name to be specified because {count} '
                 'containers are defined in this project.'.format(count=len(self.containers)))
-        containers[0].shell()
+        for container in self._containers_generator(containers=containers):
```
ellmetha: Perhaps we could write a context manager to do this when we are considering a single container. It's a small improvement (not needed now I'd say), we could add it later. :wink:

hsoft: yeah, probably.

PR #24, opened 2017-02-10
Guests / hosts / provisioners improvements

ellmetha:


nomad/container.py
```diff
@@ -306,57 +350,36 @@ def _setup_shares(self):
...
-            shareconf = {
-                'type': 'disk',
-                'source': source,
-                'path': share['dest'],
-            }
+                    self._host.give_mapped_user_access_to_share(source)
+                    # We also give these permissions to any user that was created with LXD-Nomad.
+                    for uconfig in self.options.get('users', []):
+                        username = uconfig.get('name')
+                        self._host.give_mapped_user_access_to_share(
```
hsoft: Yes, it works! In an unprivileged container and with some `users:` defined, I could properly access shares both ways. good work!

nomad/guests/base.py
```diff
@@ -0,0 +1,147 @@
...
+                break
+        return found
+
+    def add_ssh_pubkey_to_root_authorized_keys(self, pubkey):
+        """ Add a given SSH public key to the root user's authorized keys. """
+        logger.info("Adding {} to machine's authorized keys".format(pubkey))
+        self.run(['mkdir', '-p', '/root/.ssh'])
+        self.lxd_container.files.put('/root/.ssh/authorized_keys', pubkey)
+
+    def create_user(self, username, home=None):
```
hsoft: What about making the default for `home` to be `/home/{ username }` (which is the most common case) rather than no home at all? We could support `home: false` too support cases where we want to create homeless users.

nomad/guests/archlinux.py
```diff
@@ -0,0 +1,28 @@
+from .base import Guest
+
+
+class ArchLinuxGuest(Guest):
+    """ This guest can provision ArchLinux containers. """
+
+    name = 'arch'
+    barebones_packages = [
+        'openssh',
+        'python',
```
hsoft: On arch, the python 2 package is `python2`.

nomad/container.py
```diff
@@ -388,6 +411,26 @@ def _container(self):
...
         """ Returns a boolean indicating if the container has a static IP. """
         return self._container.config.get('user.nomad.static_ip') == 'true'
+
+    @property
+    def _host(self):
+        """ Returns the `Host` instance associated with the considered host.
+
+        None is returned if no guest class can be determined for the considered container. """
+        if not hasattr(self, '_container_host'):
+            host_class = next((k for k in Host.hosts if k.detect()), Host)
```
hsoft: TIL, `next()` has a default value.

nomad/guests/base.py
hsoft: Also... Wouldn't homeless users prevent shared folder permissions from being properly set up? Maybe that we shouldn't even support `home: false`!

hsoft: (or if we support it, properly document that it breaks shares permissions)

ellmetha: This method does not create homeless users because we always use the `--create-home` option (which will create a `/home/{username}` folder if we do not define another home directory).

nomad/guests/archlinux.py
ellmetha: Fixed :wink:

nomad/guests/base.py
hsoft: I read this code too fast.

PR #26, opened 2017-02-20
Add missing dependency for 'requests' version

dctremblay:
I tried to run `sudo python3 setup.py install` but I got that error message :

```
error: requests 2.4.3 is installed but requests!=2.8.0,>=2.5.2 is required
```

This patch fixes that issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
         'pylxd>=2.2,<3.0',
         'PyYAML>=3.0,<4.0',
         'voluptuous>=0.9,<1.0',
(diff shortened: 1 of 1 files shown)


hsoft:
First, you should almost never have to install python packages as `sudo`. The proper way to install lxd-nomad is with `pip3 install --user .`

Second did you try installing it with the command documented at https://lxd-nomad.readthedocs.io/en/latest/getting_started.html#install-lxd-nomad ? The `requests< dependency comes from `pylxd`. We shouldn't have to explicitly list it.

PR #32, opened 2017-02-25
Fixed #28 -- Added a command for generating a basic LXDock file

### LINKED ISSUE #28: Add an init command
An ``init`` command could be added in order to automatically create a LXDock file containing a basic configuration with instructions to get started. The container image could be set using a specific option, eg:

```
$ lxdock init
$ lxdock init --image ubuntu/xenial
```

ellmetha:


lxdock/cli/main.py
```diff
@@ -160,6 +170,29 @@ def help(self, args):
...
+            if os.path.exists(os.path.join(cwd, filename))]
+        if existing_config and not args.force:
+            raise CLIError(
+                'An existing LXDock file is already present in this directory! Using "lxdock init" '
+                'could overwrite this file. Use the -f/--force to overwrite existing LXDock files.')
+
+        # Compute the content of the LXDock file to write and write it to a lxdock.yml file.
+        init_filecontent = INIT_LXDOCK_FILE_CONTENT.format(
+            project_name=project_name, image=args.image or 'ubuntu/xenial')
+        with codecs.open('lxdock.yml', mode='w', encoding='utf-8') as fd:
```
hsoft: Why not `open()` directly?

tests/unit/cli/test_main.py
```diff
@@ -336,3 +338,78 @@ def test_project_property_works(
         project1, project2 = n.project, n.project
         assert project1 == project2
         assert from_config_mock.call_count == 1
+
```
hsoft: What about avoiding mocks and simply use a tmpdir? Mocking `os.path.exists` makes our test fragile because it would still pass even if the path passed to `os.path.exists()` was wrong.

lxdock/cli/main.py
ellmetha: Fixed.

tests/unit/cli/test_main.py
ellmetha: Well perhaps using a tmpdir is required too. Instead, I choose to mock `os.getcwd` in order to use a directory where a valid LXDock file is present.

ellmetha:
@hsoft I've updated the tests in this PR to comply with your changes in #33. :wink:

PR #35, opened 2017-03-05
docs: add "Problems?" section to "Getting Started"

hsoft:
The old readme had something about trying privileged mode when
encountering problems and this disappeared in the new docs. Considering
that `debian/jessie` is a popular image and that it doesn't work in
unprivileged mode, I think it's important to have that section to avoid
being swamped with troubleshooting issues of this type on github.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -157,3 +157,11 @@ Now from your project directory, start up your containers using the following co
   ==> test02: Provisioning container "test02"...
 
 *Congrats! You're in!*
+
+Problems?
+---------
+
+If you're having problems trying to run your container, try running them in :ref:`conf-privileged`
+mode. Many older distributions have an init system that doesn't work well with unprivileged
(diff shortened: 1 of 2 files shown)


docs/getting_started.rst
```diff
@@ -157,3 +157,11 @@ Now from your project directory, start up your containers using the following co
   ==> test02: Provisioning container "test02"...
 
 *Congrats! You're in!*
+
+Problems?
```
ellmetha: I think it could be a better solution to put this into a "Troubleshouting" section that would appear in the left menu. IMO we should try to keep the "Getting Started" section as short as possible. Watcha think?

hsoft: With only one troubleshooting tip, it seems overkill. When we reach two?

ellmetha: Alright, let's do this. :wink:

PR #47, opened 2017-03-20
Accept custom LXC configuration

elventear:
Provide a way to pass custom LXC parameters for a container, as an advanced setting.

The need for this arose from this bug:

https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/1575779

The only way I have to found to work around is to set the container to `raw.lxc: lxc.aa_profile=unconfined`.

### DIFF
--- a/docs/conf.rst
+++ b/docs/conf.rst
@@ -122,6 +122,25 @@ And then use it in your LXDock file as follows:
   image: old-ubuntu
   mode: local
 
+lxc_config
+-----------
+
+If your container needs custom configuration settings, you can use ``lxc_config`` to pass any arbitrary
+key-value pairs that you want to be assigned to the container at startup. ``lxc_config`` must be a dictionary 
+that with an arbitrary number of key-values.
+
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+  lxc_config:
+    global_key: global_value
+
+  containers:
+    - name: test01
+      test1_key: test1_value
+    - name: test02
+
 mode
 ----
 
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -259,6 +259,17 @@ class Container:
             'from image {image}'.format(name=self.lxd_name, image=self.options['image']))
         privileged = self.options.get('privileged', False)
         mode = self.options.get('mode', 'pull')
+
+        # Get user defined lxc configs
+        lxc_config = self.options.get('lxc_config', {}).copy()
+
+        # Overwrite any configuration settings with lxdock defaults
+        lxc_config.update({
+            'security.privileged': 'true' if privileged else 'false',
+            'user.lxdock.made': '1',
+            'user.lxdock.homedir': self.homedir,
+        })
+
         container_config = {
             'name': self.lxd_name,
             'source': {
@@ -280,12 +291,9 @@ class Container:
                            else ''),
                 'type': 'image',
             },
-            'config': {
-                'security.privileged': 'true' if privileged else 'false',
-                'user.lxdock.made': '1',
-                'user.lxdock.homedir': self.homedir,
-            },
+            'config': lxc_config,
         }
+
         try:
             return self.client.containers.create(container_config, wait=True)
         except LXDAPIException as e:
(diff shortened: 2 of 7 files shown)


hsoft:
Thanks for this PR @elventear, this looks handy. Would you mind adding documentation and tests along with this?

Also, I'm wondering if the `s` suffix is appropriate in `lxc_configs`. After all, the `lxc` command line always refer to `config`, not `configs`.

lxdock/container.py
```diff
@@ -242,6 +242,17 @@ def _get_container(self, create=True):
             'from image {image}'.format(name=self.lxd_name, image=self.options['image']))
         privileged = self.options.get('privileged', False)
         mode = self.options.get('mode', 'pull')
+
+        # Get user defined lxc configs
+        lxc_config = self.options.get('lxc_configs', {}).copy()
+
+        # Overwrite any configuration settings with lxcdock defaults
```
ellmetha: `lxdock`, not `lxcdock`. :smile:

elventear:
@hsoft I'd gladly add some tests, I looked into it but this part of the code does not seem to be tested directly in the unit tests. At least the configuration dictionary that gets passed to the client. I would appreciate pointers on how to test this.

Also, no sure if we want to test this via integration too.

The reason I went with `lxc_configs` is because it allows to pass multiple configuration keys at once. `lxc` only allows one at a time, and to do several you would need to call it multiple times. But if you would prefer to change it, I am ok naming it `lxc_config`.

ellmetha:
Perhaps it would make more sense to have a `lxc_configs` option if we are considering a list of key/value configuration options. Otherwise I would go with the `lxc_config` option name.

I think you should add a test demonstrating that it is indeed possible to add a custom lxc option to a container. It could be an integration test. Also, you'll want to add this new option to the [LXDock file reference documentation](https://github.com/lxdock/lxdock/blob/master/docs/conf.rst).

tests/integration/test_project.py
```diff
@@ -14,6 +15,14 @@
 FIXTURE_ROOT = os.path.join(os.path.dirname(__file__), 'fixtures')
 
 
+@contextlib.contextmanager
```
hsoft: Your tests are looking good, thanks! I don't mean to be nitpicking, but I don't think you'll be needing this context manager for your tests. As long as your container names start with `lxdock-test-` (correct me if I'm wrong @ellmetha ), cleanup will be automatically taken care of at https://github.com/lxdock/lxdock/blob/bd9d01152e08dc828d8e87de8ef267e35106fd94/lxdock/test/testcases.py#L17

ellmetha: @elventear you can use the `containername` helper to construct your test container names (see: https://github.com/lxdock/lxdock/blob/bd9d01152e08dc828d8e87de8ef267e35106fd94/lxdock/test/testcases.py#L40).

tests/integration/test_project.py
```diff
@@ -156,3 +165,14 @@ def test_can_open_a_shell_for_a_specific_container(self, mocked_call, persistent
         assert mocked_call.call_count == 1
         assert mocked_call.call_args[0][0] == \
             'lxc exec {} -- su -m root'.format(persistent_container.lxd_name)
+
+    def test_set_lxc_config(self):
```
ellmetha: I think that you should've placed this test in `integration/test_container.py` because it is related to a feature implemented in the `Container` class. :wink:

docs/conf.rst
```diff
@@ -108,6 +108,25 @@ And then use it in your LXDock file as follows:
   image: old-ubuntu
   mode: local
 
+lxc_configs
```
ellmetha: IMO `lxc_config` would make more sense because we are considering a dictionary.

elventear:
Sorry for the delay, life has kept me busy and just recently found time to finalize this. I have tried to address all comments. I have rebased and squashed the PR.

ellmetha:
Merged. Thanks @elventear! :wink:

docs/conf.rst
```diff
@@ -108,6 +108,25 @@ And then use it in your LXDock file as follows:
...
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+  lxc_configs:
+    global_key: global_value
+
+  containers:
+    - name: test01
+      test1_key: test1_value
```
elventear: It slipped through the review. It should be removed.

ellmetha: Alright! Thanks for reminding this to us!

PR #50, opened 2017-03-26
shell: add a way to set environment variables

hsoft:
Add a new config dictionary in the `shell` key: `environment`. All
values in this dictionary are passed to LXD's config under the
"environment.*" config type.

We set this config when `shell` is called, which is a bit at odds with
the rest of the container config, which is usually set at creation time.
We do this so that a user who fiddles with her `lxdock.yml` file doesn't
have to restart/recreate her container to benefit from the change she's
just made. They will be effective on the next `lxdock shell` call.

### DIFF
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -1,4 +1,4 @@
-from voluptuous import All, Any, In, IsDir, Length, Required, Schema, Url
+from voluptuous import All, Any, Extra, In, IsDir, Length, Required, Schema, Url
 
 from ..provisioners import Provisioner
 
@@ -21,6 +21,7 @@ _top_level_and_containers_common_options = {
     'shell': {
         'user': str,
         'home': str,
+        'environment': {Extra: str},
     },
     'users': [{
         # Usernames max length is set 32 characters according to useradd's man page.
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -6,6 +6,7 @@ from .validators import Hostname, LXDIdentifier
 
 
 _top_level_and_containers_common_options = {
+    'environment': {Extra: Coerce(str)},
     'hostnames': [Hostname(), ],
     'image': str,
     'mode': In(['local', 'pull', ]),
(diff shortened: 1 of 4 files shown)


hsoft:
hum... tests were working locally. Maybe it's because I reuse the same `dummy` container name. Will try to fix later.

lxdock/conf/schema.py
```diff
@@ -21,6 +21,7 @@
     'shell': {
         'user': str,
         'home': str,
+        'environment': {Extra: str},
```
ellmetha: I think that this rule is a bit too restrictive. For example, during my tests I tried to define environment variables as follows:

```
shell:
  environment:
    LC_ALL: en_US.utf8
    TEST_VAR: 42
```

This fails because the `TEST_VAR` variable is converted to an integer after YML loading, and we are expecting a string. :wink:

ellmetha:
I understand the need for such an option. But shouldn't we add a general `environment` option (comparable to the docker-compose's environment option ; see: https://docs.docker.com/compose/compose-file/#environment)? Environment variables could be useful out of the scope of the `shell` command. If we were to support such an option, I think it would make sense to:

* ensure that the environment variables are set during the `up` action or the `provision` action
* are updated during the `shell` action

hsoft:
The test I've made indicate that the trick I've used, setting `environment.*` config only affects `lxc exec`. For example, if you set `environment.FOO` to `bar`, `echo $FOO` will give you `bar` under `lxc exec`, but if you ssh into the same container, `echo $FOO` returns nothing.

Given this, and unless I'm missing something, moving that config tweaking routine to `lxdock up` would be useless because it only affects `lxdock shell` anyway.

If we want to support "machine-wide" environment variable setting, we will need another way to do so (`/etc/profile` mangling? but that would probably be guest-OS-specific) which is likely to be more complicated than the solution proposed here.

ellmetha:
So these environment variables are only available if we use `lxc exec`. I guess that environment variables are also available if we use the `container.execute` method provided by pylxd (which seems to use the `exec` method internally) - and we are using this method in order to run commands (eg. to install packages). We don't have a `shell` provisioner yet but such a provisioner could benefit from a global `environment` option (because the command would be executes on the guests using the exec method).

hsoft:
I ran some tests by using your code from #56 and yes, `environment.*` affects shell provisioning, so it should indeed be set on `lxdock up`, `lxdock provision` and `lxdock shell`.

PR #53, opened 2017-03-27
Improve user's home directory assignment

isubas:
If a shell user defined without a home directory setting, assign home directory to /home/user (unless root).

### DIFF
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -105,7 +105,11 @@ class Container:
         shelluser = username or shellcfg.get('user')
         if shelluser:
             # This part is the result of quite a bit of `su` args trial-and-error.
-            shellhome = shellcfg.get('home') if not username else None
+            shellhome = None
+            if not username:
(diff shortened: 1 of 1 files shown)


hsoft:
Thanks for your PR @isubas . This is indeed something that has to be improved. The problem with your proposal, however, is that it only works in the case where we're logging into the user defined in the `lxdock.yml` config. If we do something like `lxdock shell -u anotheruser`, we will not end up in `/home/anotheruser`. It would be better if we implemented the fix for this use case as well.

hsoft:
It look good to me, thanks. Would you mind adding a test? Tests might look like a big chunk to grok at first, but if you inspire yourself from something like https://github.com/lxdock/lxdock/blob/master/tests/integration/test_container.py#L86 , it's almost a matter of copy/pasting.

hsoft:
Closing PR due to inactivity. Will re-open if activity resume.

PR #54, opened 2017-03-29
Add support for ansible options related to ansible-vault

ellmetha:


hsoft:
What about simply allowing arbitrary command line args? I don't see the gain in copying ansible's command line args in our yaml schema.

ellmetha:
I don't think that this could work if the `lxdock` command is used from subfolders. Because you would add something like `raw_arguments: --vault-password-file secrets/mypass` to your LXDock file. Otherwise, this is mostly for convenience. The gain is to make the LXDock configuration easier (and more readable) for the options that are the most commonly used (that said this doesn't prevent us to support a `raw_arguments` option if necessary).

PR #55, opened 2017-03-30
Make "barebone" packages be installed for ansible only

hsoft:
Those packages were always a misnomer. They're not really "barebone",
they're required only for ansible to be able to run properly. With new
provisioners coming in soon, it becomes inefficient to install those
packages on *every* container, even those not running ansible.

Moreover, it has an immediate side effect: it makes some integration
tests run much faster.

### DIFF
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -80,12 +80,16 @@ class Container:
         if barebone is None:  # None == only if the container isn't provisioned.
             barebone = not self.is_provisioned
 
+        provisioning_steps = self.options.get('provisioning', [])
+
         if barebone:
             self._perform_barebones_setup()
+            if any(ps['type'].lower() == 'ansible' for ps in provisioning_steps):
+                self._guest.install_ansible_packages()
 
         logger.info('Provisioning container "{name}"...'.format(name=self.name))
 
-        for provisioning_item in self.options.get('provisioning', []):
+        for provisioning_item in provisioning_steps:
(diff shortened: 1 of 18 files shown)


hsoft:
@ellmetha I don't understand this test failure. Could it be that our tests are flaky? When I look at https://github.com/lxdock/lxdock/blob/master/lxdock/test/testcases.py#L30 , I see a similar `teardown_class` and `teardown_method`. Maybe that in some cases, pytest runs these in parallel, causing a race condition making our container being in `test_containers`, but then being deleted before we have the chance to run `delete()`?

ellmetha:
I agree that we have install the packages required by provisioning tools only if we are using the considered provisioners. But with this PR the installation of the dependencies required for Ansible provisioning is tightly coupled with the container main class and the guests. I don't think that this is the best way to achieve the task at hand. In my opinion `Provisioner` subclasses should be able to install the packages they need themselves. I think that we have to ensure that `Host` or `Guest` subclasses remain loosely coupled to the way provisioners work. :wink:

hsoft:
We're talking about an incremental improvement that makes Travis tests go from 12 min to 8 min and moves around code that was *already* tighly coupled.

Anyway, if you insist on doing it the "right way" immediately, I'll let you do it.

ellmetha:
I'll merge this PR because it addresses a real problem. In the next few weeks I should be working on improving the separation of concern of hosts/guests/provisioners. :wink:

PR #56, opened 2017-03-30
Add shell provisioner

ellmetha:
Depends on #54. :wink:

### DIFF
--- a/lxdock/provisioners/base.py
+++ b/lxdock/provisioners/base.py
@@ -94,4 +94,4 @@ class Provisioner(with_metaclass(_ProvisionerBase)):
 
     def homedir_expanded_path(self, path):
         """ Expands the considered path with the path of the home homedir if applicable. """
-        return os.path.join(self.homedir, path) if not os.path.isabs(path) else path
+        return os.path.join(self.homedir, path)
--- a/lxdock/provisioners/base.py
+++ b/lxdock/provisioners/base.py
@@ -92,6 +92,6 @@ class Provisioner(with_metaclass(_ProvisionerBase)):
     # HELPER METHODS #
     ##################
 
-    def homedir_expanded_path(self, relative_path):
-        """ Expands the considered path with the absolute path of the home homedir. """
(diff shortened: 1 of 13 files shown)


hsoft:
We're again victims of flaky tests. See https://github.com/lxdock/lxdock/pull/55#issuecomment-290278050 . We have to find some solution to this.

lxdock/provisioners/base.py
```diff
@@ -91,6 +92,6 @@ def provision(self):
     # HELPER METHODS #
     ##################
 
-    def homedir_expanded_path(self, relative_path):
-        """ Expands the considered path with the absolute path of the home homedir. """
-        return os.path.join(self.homedir, relative_path)
+    def homedir_expanded_path(self, path):
+        """ Expands the considered path with the path of the home homedir if applicable. """
+        return os.path.join(self.homedir, path) if not os.path.isabs(path) else path
```
hsoft: In your initial commit, I wanted to suggest that you do that, but it turns out that it's not needed: https://docs.python.org/3/library/os.path.html#os.path.join:

`If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.`

So I said nothing.

ellmetha: Fixed! :wink:

hsoft:
Isn't the "steps" concept redundant? We can already organize our provisioning items in ordered steps. What do we gain by having sub-steps?

hsoft:
I also notice that piping (`|`, `>`, `<`) doesn't work for inline shell. I'm not quite sure about the difficulty involved in supporting that, but we should at least mention the limitation in the doc.

ellmetha:
@hsoft you're right, ``steps`` are not necessary. I've changed this behaviour. I've also added a note related to the use of common shell patterns (like `>`/`|`/...) to the documentation. It turns out that `lxc exec` doesn't use a shell internally so the problem can be overcame by specifying commands such as `sh -c 'ls -l > /tmp/test`.

PR #57, opened 2017-03-31
tests: remove spurious class-level teardown

hsoft:
We've been having multiple cases of flaky tests failures from Travis
lately and they always come from the class-level teardown. Because I
can't figure why we even have it here (in case we ever, hum, create some
test containers *after* the last test method has run?) and because
removing it didn't result in tests leaking out test containers on my
machine, I therefore propose to remove it and make our tests stop being
flaky.

### DIFF
--- a/lxdock/test/testcases.py
+++ b/lxdock/test/testcases.py
@@ -26,10 +26,6 @@ def _remove_test_containers(client=None):
 class LXDTestCase:
     """ Base test class used to run integration tests that require the use of LXD. """
 
-    @classmethod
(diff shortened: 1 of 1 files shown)


hsoft:
Ok, now we know that it doesn't change anything :)

hsoft:
Eureka! https://github.com/lxc/pylxd/blob/master/pylxd/models/_model.py#L178 it's so clear now: we have to delete with `wait=True`! If we don't do that, **of course** we put ourselves in a situation of a possible race condition.

PR #58, opened 2017-03-31
Ensured guests are loosely coupled with provisioners

ellmetha:
Depends on #56.

### DIFF
--- a/lxdock/provisioners/ansible.py
+++ b/lxdock/provisioners/ansible.py
@@ -18,12 +18,12 @@ class AnsibleProvisioner(Provisioner):
     guest_required_packages_alpine = ['openssh', 'python', ]
     guest_required_packages_archlinux = ['openssh', 'python2', ]
     guest_required_packages_centos = ['openssh-server', 'python', ]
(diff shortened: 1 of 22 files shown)


lxdock/provisioners/ansible.py
```diff
@@ -14,6 +14,17 @@ class AnsibleProvisioner(Provisioner):
     """ Allows to perform provisioning operations using Ansible. """
 
     name = 'ansible'
+
+    guest_required_packages_alpine = ['openssh', 'python', ]
+    guest_required_packages_archlinux = ['openssh', 'python2', ]
+    guest_required_packages_centos = ['openssh-server', 'python', ]
+    guest_required_packages_debian = ['openssh-server', 'python', ]
```
hsoft: Technically, it's enough, but without `python-apt` and `aptitude`, we can't run the `apt` action, which is used pretty much all the time: http://docs.ansible.com/ansible/apt_module.html#requirements-on-host-that-executes-module

ellmetha: @hsoft yep, I've just added these missing packages to the list. :wink:

PR #65, opened 2017-04-28
Added support for variable substitution in LXDock files

ellmetha:


lxdock/conf/config.py
```diff
@@ -75,11 +77,47 @@ def from_base_dir(cls, base_dir='.'):
...
+        """
+        # Builds a dictionary of variables that will be used to perform variable substitution in the
+        # config values.
+        mapping = {}
+        mapping.update(os.environ)
+
+        # Fetches variables that could be defined in a .env file and inserts them in the mapping.
+        env_filepath = os.path.join(self.homedir, '.env')
+        if os.path.exists(env_filepath):
+            mapping.update({k: v for k, v in self._get_dotenv_items(env_filepath)})
```
hsoft: What about `dict(self._get_dotenv_items(env_filepath))`?

lxdock/conf/config.py
```diff
@@ -115,6 +153,27 @@ def _get_container_config_dict(self, container_dict):
         del container_config['containers']
         return container_config
 
+    def _get_dotenv_items(self, env_filepath):
```
hsoft: Is the `.env` format supposed to be a standard of some sort? According to rules at https://github.com/bkeepers/dotenv, we're missing a few cases. People using `.env` file that aren't specific to lxdock will probably expect those to be parsed as if the file was interpreted in `bash` and this function will never stop growing more and more complex.

Couldn't we simply `source` the file?

lxdock/conf/interpolation.py
```diff
@@ -0,0 +1,25 @@
...
+    def interpolate(value):
+        if isinstance(value, str):
+            return Template(value).substitute(**mapping)
+        elif isinstance(value, dict):
+            return {k: interpolate(v) for k, v in value.items()}
+        elif isinstance(value, (list, tuple)):
+            return [interpolate(v) for v in value]
+        return value
+
+    return dict(reduce(
```
hsoft: Why the `reduce()`? Isn't this a typical case for a dict comprehension? This line seems overly complex to me.

lxdock/conf/config.py
ellmetha: Fixed.

lxdock/conf/interpolation.py
ellmetha: Fixed.

lxdock/conf/config.py
ellmetha: I understand your point. That said other Python tools (eg. docker-compose) seem to use similar techniques to process ``.env`` files, which seems fine IMO. If we were to ``source`` these ``.env`` files, I guess this would mean we would have to do something like ``source .env && env`` and then parse the output of this command. But this would also involve declaring vars using `export` in the `.env` file if I'm correct.

hsoft: Yeah, that seems tricky. Maybe we could simply add a `printenv` call at the end of the `.env` script and parse *that* instead. `printenv` would give us something that is easier to parse.

hsoft: ... but then it's probably a dead end. After all, they developed a full-featured library at https://github.com/bkeepers/dotenv . It seems like we have 2 basic choices:

1. Develop a full featured dotenv-parsing library. A quick google search only reveal libraries that don't go much further than your parsing routine.
2. Accept and document sub-standard parsing.

The problem with going with the second option will be that this routine risk being gradually "improved" through various PR to make it more "bash-like" and will end up being bigger than we'd like.

ellmetha: Or we could use something like https://github.com/theskumar/python-dotenv and add it to the project's dependencies.

hsoft: It's probably the best option (if we really need this feature).

ellmetha: Alright. I've just updated the branch to use python-dotenv. :wink:

robvdl:
This broke things for me because, now I suddenly have to replace $ in password hashes with $$.

I created a bug for this, as I don't think this is very nice, as soon as we upgrade from lxdock 0.2 to 0.3 things will break.

Why can't we use jinja2 for this, just like ansible does already for example.  Since jinja2 uses {{var}} this is way more robust and won't require $'s to be escaped.

PR #67, opened 2017-05-03
Fixed #61: add support for passing `-c command_line` to lxdock shell

### LINKED ISSUE #61: Add a `run` command to the client
To run commands within the container, a cli command such as `lxdock run --user <user> "command"` would be very useful to have.

The implementation of `lxdock shell` can almost support this already, would just need to append `-c <command>` to the `lxc exec <container> -- su - <user> ` command.

lingxiaoyang:
After discussing with @ellmetha, #61 is implemented as an additional parameter `-c`:

```
lxdock shell [arguments] [name] [-c command_line]
```

The parameter `-c` should be placed at the last and it takes all remaining arguments as `command_line`. With a similar behavior of `docker exec` and `lxc exec`, the arguments are passed literally and not evaluated further by shell syntax:

```
$ docker exec 12dc5e75e5a6 echo "he re\"s" -u '$PATH'
he re"s -u $PATH

$ lxc exec test-linux-containers-gentoo-current-13382216 -- echo "he re\"s" -u '$PATH'
he re"s -u $PATH

$ lxdock shell gentoo-current -u localuser -c echo "he re\"s" -u '$PATH'
he re"s -u $PATH
```

The added dependency `shlex` in `lxdock/container.py` has previously existed in `lxdock/provisioners/shell.py`, so nothing changes regarding dependencies.

The bash and zsh completions are also updated. For bash, `-c, --command` is added besides `-u, --username`. But zsh is able to take all remaining arguments:
image

Related docs are updated.

### DIFF
--- a/docs/cli/shell.rst
+++ b/docs/cli/shell.rst
@@ -1,18 +1,20 @@
 lxdock shell
 ============
 
-**Command:** ``lxdock shell [arguments] [name]``
+**Command:** ``lxdock shell [name] [arguments]``
 
 This command can be used to open an interactive shell inside one of your containers.
+If ``-c`` is specified, execute the command line instead of opening the shell.
 
 By default, that shell logins as ``root`` unless your LXDock config specifies another user
-in its ``shell:`` option. In all cases, the ``--user`` command line overrides everything.
+in its ``shell`` option. In all cases, the ``--user`` command line overrides everything.
 
 Options
 -------
 
 * ``[name]`` - a container name
 * ``-u, --user <username>`` - user to login as
+* ``-c, --command <command_line>`` - command to be executed in the shell
 
 Examples
 --------
@@ -21,3 +23,8 @@ Examples
 
   $ lxdock shell mycontainer       # opens a shell into the "mycontainer" container
   $ lxdock shell -u root           # opens a shell as root, regardless of our config
+  $ lxdock shell mycontainer -c echo HELLO      # executes "echo HELLO" in "mycontainer"
+  $ lxdock shell mycontainer -c echo '$PATH'    # executes "echo '$PATH'" in "mycontainer"
+
+
+For the last example, you will see "$PATH" as-is. It is not evaluated as a variable.
--- a/lxdock/cli/main.py
+++ b/lxdock/cli/main.py
@@ -74,11 +74,16 @@ class LXDock:
 
         # Creates the 'shell' action.
         self._parsers['shell'] = subparsers.add_parser(
-            'shell', help='Open a shell in a container.',
-            description='Open an interactive shell inside a specific container.')
+            'shell', help='Open a shell or execute a command in a container.',
+            description='Open an interactive shell inside a specific container. If a command is '
+                        'specified, execute the command instead.',
+            usage='lxdock shell [-h] [-u USERNAME] [name] [-c ...]')
         self._parsers['shell'].add_argument('name', nargs='?', help='Container name.')
         self._parsers['shell'].add_argument(
             '-u', '--username', help='Username to login as.')
+        self._parsers['shell'].add_argument(
+            '-c', '--command', nargs=argparse.REMAINDER, dest='cmd_args',
+            help='Command to be executed.')
 
         # Creates the 'status' action.
         self._parsers['status'] = subparsers.add_parser(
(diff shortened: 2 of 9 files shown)


hsoft:
> The added dependency shlex

FYI: [shlex](https://docs.python.org/3/library/shlex.html) is a builtin :)

lxdock/cli/main.py
```diff
@@ -74,11 +74,16 @@ def __init__(self, argv=None):
...
-            description='Open an interactive shell inside a specific container.')
+            'shell', help='Open a shell or execute a command in a container.',
+            description='Open an interactive shell inside a specific container. If a command is '
+                        'specified, execute the command instead.',
+            usage='lxdock shell [-h] [-u USERNAME] [name] [-c ...]')
         self._parsers['shell'].add_argument('name', nargs='?', help='Container name.')
         self._parsers['shell'].add_argument(
             '-u', '--username', help='Username to login as.')
+        self._parsers['shell'].add_argument(
+            '-c', '--command', nargs=argparse.REMAINDER, dest='commands',
```
hsoft: `commands` seems like a misnomer to me. We always have a single command, not multiple commands. What about `command_args`?

hsoft:
This PR looks good to me, but behavior-wise, we'll need a fix for a problem I've noticed: the command isn't properly stopped on CTRL-C.

For example, if I run something like Django's "manage runserver" through `lxdock shell -c`, it will work and CTRL-C will bring me back to the prompt, but the "runserver" process will keep running. I've tried it with a direct `lxc exec` command and it doesn't have the same problem.

docs/cli/shell.rst
```diff
@@ -1,9 +1,10 @@
 lxdock shell
 ============
 
-**Command:** ``lxdock shell [arguments] [name]``
+**Command:** ``lxdock shell [arguments] [name] [-c command_line]``
```
ellmetha: I think that ``lxdock shell [name] [arguments]`` would be more appropriate. Command arguments are not explicitly defined in this line in other command reference pages, so It'd be best to try to remain consistent here too. 😉

lxdock/cli/main.py
lingxiaoyang: @hsoft  Yes, you're right. We're using `Host.run(cmd_args)` and `Guest.run(cmd_args)` in the code. I'll change to `cmd_args`. Thank you!

lingxiaoyang:
@hsoft  I confirm this behavior. I'll try to see why it happens...

lingxiaoyang:
@hsoft  I did more testing on this:

```bash
# Ctrl-C worked fine
lxc exec modele-default-14288073 --env HOME=/opt/demo_django -- ./develop.sh

# Ctrl-C didn't work
lxc exec modele-default-14288073 --env HOME=/opt/demo_django -- su -m demo_django -c ./develop.sh
```

So it seems that the problem was caused by `su -c`. [Here](http://sethmiller.org/it/su-forking-and-the-incorrect-trapping-of-sigint-ctrl-c/) is the only article that I found explaining this issue. 

In 9d0c01a, I store the command line as a bash script and use `su -s shell_file` to let SIGINT work. The article above suggests the use of `--session-command` but it didn't seem to me that Ubuntu 16.04 supported `su --session-command`. So I have to use `su -s` as it's more supported.

hsoft:
Good find, thanks for the research on the subject! Testing it right away.

lxdock/container.py
```diff
@@ -37,6 +40,9 @@ class Container:
     # The default image server that will be used to pull images in "pull" mode.
     _default_image_server = 'https://images.linuxcontainers.org'
 
+    # The default path for storing the command to execute during `lxdock shell`.
+    _guest_shell_script_file = '/.lxdock.d/shell_cmd.sh'
```
hsoft: This will probably bite us at some point (having a single hadcoded path for all shell commands). We might want to do the right thing right away and go with a `mktemp` approach.

hsoft:
Works well, merging. My inline `mktemp` comment was just for the record. We'll probably want to build ourselves, at some point, a `Guest` utility to run arbitrary scripts that will use `mktemp` properly. When we do, we'll piggy-back on it in `lxdock shell -c`.

Issue #37, opened 2017-03-08
Add support for new provisioning tools


ellmetha:
* chef
* puppet
* salt
* shell provisioning?
* ...

luisfaceira:
I would differentiate and prioritize shell provisioning, since its likely the easiest one and in practice with a shell provisioner one can run almost any other provisioner.

jaylenw:
Hey, this may be a noob question but how can I have lxdock run a simple bash script upon provision?

Thank you.

hsoft:
@jaylenw I'm glad you asked! it's a new feature of the newly-released v0.2 :)

See https://lxdock.readthedocs.io/en/stable/provisioners/shell.html

jaylenw:
@hsoft Cool! thank you!

erozqba:
@ellmetha you could remove "puppet" from the list now, thanks to @lingxiaoyang :+1:

PR #69, opened 2017-05-04
New: Puppet provisioning (experimental support). Refs #37

lingxiaoyang:
This pull request adds puppet provisioning support for LXDock, tested with [this lxdock.yml and site.pp](https://gist.github.com/lingxiaoyang/83dc9291073876342bbb6f54488f685f). 

New module: `lxdock.provisioners.puppet.PuppetProvisioner`.

It is implemented with the intention to mimic Vagrant's Puppet provisioning. The options are taken from [Vagrant doc](https://www.vagrantup.com/docs/provisioning/puppet_apply.html#options) and the usage of `puppet apply` command is taken from [Vagrant implementation](https://github.com/mitchellh/vagrant/blob/9c299a2a357fcf87f356bb9d56e18a037a53d138/plugins/provisioners/puppet/provisioner/puppet.rb#L173).

New public methods: `Guest.copy_file(host_path, guest_path)` and `Guest.copy_directory(host_path, guest_path)` (uses `pathlib` that exists since Python 3.4)

### DIFF
--- a/docs/provisioners/puppet.rst
+++ b/docs/provisioners/puppet.rst
@@ -1,0 +1,93 @@
+#######
+Puppet
+#######
+
+LXDock provides built-in support for `Puppet`_ provisioning.
+
+.. _Puppet: https://puppet.com/
+
+
+Distribution Support
+--------------------
+
+Puppet provisioning requires the executable ``puppet`` in the guest container. For ArchLinux,
+Debian, Fedora and Ubuntu guests, LXDock tries to install the ``puppet`` package during
+``lxdock up``.
+
+However, the automatic installation may fail, it may install an old version of Puppet, or you
+may also use another Linux distribution. In these cases, you may see an error message at
+the beginning of ``lxdock provision`` or during the provisioning (due to version mismatch).
+You will have to check out the proper documentation for your Linux distribution and Puppet version,
+and add a shell provisioner in LXDock file before the Puppet provisioner. Here is an example for
+CentOS 7:
+
+.. code-block:: yaml
+
+  name: centos-7
+  image: centos/7
+
+  provisioning:
+    - type: shell
+      inline: sh -c 'rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-el-7.noarch.rpm && yum -y install puppet'
+    - type: puppet
+      [...] # Puppet options
+
+By default, LXDock runs ``which puppet`` to find the executable. You may override this behavior
+by providing ``binary_path`` option and LXDock will then find ``puppet`` executable under that path.
+
+
+Usage
+-----
+
+Add a ``puppet`` provisioning operation to your LXDock file as follows:
+
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+
+  provisioning:
+    - type: puppet
+      manifests_path: manifests
+      manifest_file: site.pp
+      module_path: modules
+      hiera_config_path: hiera.yaml
+      facter:
+          role: app
+          domain_name: app.example.com
+      options: "--verbose --debug"
+
+
+Puppet provisioning can be run in "manifest" mode or "environment" mode by setting ``manifests_path``
+or ``environment_path``, respectively. If ``manifest_file`` is not provided in "manifest" mode, it is
+given the default value ``default.pp``. If ``environment`` is not provided in "environment" mode, it
+is given ``production`` as the default value.
+
+If none of ``manifests_path`` and ``environment_path`` are given, LXDock assumes "manifest" mode and
+set ``manifests_path`` to ``manifests``. During the validation of an LXDock file, the existence of
+``manifests_path/manifest_file`` or ``environment_path/environment`` is checked.
+
+
+Options
+-------
+
+LXDock's Puppet provisioning is expected to reuse the files and configurations for a Vagrant project.
+This is still experimental, so if it doesn't work for your case, please feel free to create a GitHub
+issue!
+
+Here are the options that LXDock has supported:
+
+- ``binary_path``
+- ``facter``
+- ``hiera_config_path``
+- ``manifest_file``
+- ``manifests_path``
+- ``module_path``
+- ``environment``
+- ``environment_path``
+- ``environment_variables``
+- ``options``: LXDock takes a single string of space-separated options, instead of an array of strings.
+
+Please reference: `Vagrant docs`_
+
+.. _Vagrant docs: https://www.vagrantup.com/docs/provisioning/puppet_apply.html#options
--- a/lxdock/provisioners/puppet.py
+++ b/lxdock/provisioners/puppet.py
@@ -1,0 +1,201 @@
+import logging
+import shlex
+from pathlib import Path, PurePosixPath
+
+from voluptuous import All, IsDir, IsFile
+
+from ..exceptions import ProvisionFailed
+from .base import Provisioner
+
+
+__all__ = ('PuppetProvisioner', )
+
+logger = logging.getLogger(__name__)
+
+
+def finalize_options(options):
+    # Refs Vagrant code:
+    # https://github.com/mitchellh/vagrant/blob/9c299a2a357fcf87f356bb9d56e18a037a53d138/
+    #         plugins/provisioners/puppet/config/puppet.rb#L58
+    if options.get('environment_path') is None and options.get('manifests_path') is None:
+        logger.warn("environment_path and manifests_path are both unset, "
+                    "assuming manifests mode and manifests_path as 'manifests'...")
+        options['manifests_path'] = 'manifests'
+
+    if options.get('environment_path') is None:
+        if options.get('manifest_file') is None:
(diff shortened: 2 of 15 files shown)


lxdock/provisioners/ansible.py
```diff
@@ -40,10 +40,31 @@ def provision(self):
...
     def setup_guest_alpine(self):
-        # On alpine guests we have to ensure that ssd is started!
+        # On alpine guests we have to ensure that sshd is started!
         self.guest.run(['rc-update', 'add', 'sshd'])
         self.guest.run(['/etc/init.d/sshd', 'start'])
 
+    def setup_guest_arch(self):
+        # On archlinux guests we have to ensure that sshd is started!
+        # https://bbs.archlinux.org/viewtopic.php?id=205367
+        self.guest.run(['systemctl', 'enable', 'sshd.socket', '--now'])
```
lingxiaoyang: Question: I am repeating this "ensure sshd" code in both `ansible.py` and `puppet.py`. Is there a better way to do this? For example, I was also thinking to put it as `ArchLinuxGuest.ensure_sshd()`.

lxdock/provisioners/puppet.py
```diff
@@ -0,0 +1,122 @@
...
+    def provision(self):
+        """ Performs the provisioning operations using puppet. """
+        ip = get_ipv4_ip(self.guest.lxd_container)
+
+        # Disable known hosts check
+        ssh_options = ['-q', '-o', 'StrictHostKeyChecking=no', '-o', 'UserKnownHostsFile=/dev/null']
+
+        # Copy manifest file
+        self.host.run(['scp'] + ssh_options + [
+            self.options['manifest_file'], 'root@{}:{}'.format(ip, self._guest_manifest_file)])
```
lingxiaoyang: Question: is there a better way to copy files/folders?

lxdock/provisioners/puppet.py
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    _guest_manifest_file = '/tmp/site.pp'
+    _guest_module_path = '/tmp/puppet-modules'
+    _guest_environment_path = '/tmp/puppet-environment'
+
+    def provision(self):
+        """ Performs the provisioning operations using puppet. """
+        ip = get_ipv4_ip(self.guest.lxd_container)
+
```
lingxiaoyang: Note to myself: check if `puppet` exists before proceeding. If it doesn't exist, raise an error to let user know.

lxdock/provisioners/puppet.py
hsoft: Yes. See https://github.com/lxdock/lxdock/blob/master/lxdock/network.py#L106

lxdock/provisioners/ansible.py
hsoft: I think it would be a good idea, yes.

lxdock/provisioners/puppet.py
lingxiaoyang: @hsoft Thank you!

hsoft:
This looks well written to me, good job! I haven't tested all the options because I don't have a non-trivial puppet config handy here. @lingxiaoyang, was this code tested with non-trivial puppet provisioning? If yes, I think we can merge as-is.

lingxiaoyang:
@hsoft  I'll test further.

docs/provisioners/puppet.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Append a ``puppet`` provisioning operation to your LXDock file as follows:
+
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+
+  provisioning:
+    - type: ansible
```
ganto: Looks like here is still a typo

lingxiaoyang: @ganto  Please advise?

ganto:
Shouldn't this be `type: puppet`?

lingxiaoyang:
@ganto  Haha, you are right. I didn't notice...

Thank you!

docs/provisioners/puppet.rst
lingxiaoyang: Fixed in 06b87e8.

lingxiaoyang:
Reminder to myself - TODO: non-trivial puppet provisioning test.

lingxiaoyang:
Problems found when testing this puppet recipe of LAMP setting: https://github.com/hollsk/puppet-lamp

1. Need to support `manifests_path`;
2. Folder copy is too slow when there are many files, particularly the module folder.

lingxiaoyang:
Finally updated in 65dcb23!

What I have changed since the previous review:
1. After testing with a few Vagrant provisioning setups, I finally chose to copy the entire logic of Vagrant counterpart. In module [`lxdock.provisioners.puppet`](https://github.com/lxdock/lxdock/pull/69/commits/65dcb23c4cc08426779af363253c48166a436825#diff-7652a936db8ff9a996b029c58583eff7), you will see more references to Vagrant code.
2. `Guest.copy_directory` is now implemented by creating/extracting a tar file, because recursively calling `Guest.copy_file` is way too slow.
3. The doc is rewritten.

Other changes not directly related to Puppet provisioning itself (I can split them into multiple commits, if needed):
1. Add `lxdock.conf.schema.get_schema()` for testing purposes.
2. Improve the autodiscovery of provisioner schemas:
   1. support composite voluptuous schema (`Any`, `All`, ...) to be defined in provisioner class.
   2. prevent "fall-through" error message when the provisioner schema is not satisfied:
         - Previously, if the target schema (uniquely determined by "type") is not satisfied, the config will continue to be compared with remaining schemas and eventually LXDock throws an error message that is totally unrelated to target schema (because of `Any`).
3. Do the bare bone setups for the provisioners before calling any `.provision()` method, to prevent
   package clashing if the user defines other provisioners that prepare required packages for a
   later provisioner. Let's ensure that user-defined operations always override LXDock's default.

I have tested with:
- https://github.com/hollsk/puppet-lamp: "manifest" mode + Puppet modules
- https://github.com/mikespokefire/puppet-hiera-test: "manifest" mode + modules + hiera config + facter
- https://github.com/eddumelendez/vagrant-puppet-examples/tree/master/jenkins and https://github.com/eddumelendez/vagrant-puppet-examples/tree/master/mongodb: "manifest" mode + modules
- https://gitlab.berlinonline.net/puppets/deploy-template: "environment" mode + binary path + facter + env variables. `puppet apply` was successfully kicked off (the environment was correctly found) but this repo had too many dependencies so the provisioning didn't go through. 

I didn't find more valid Vagrant projects that use "environment" mode. That's why I put "experimental support" in docs as it still needs improvement, but I think it's fairly usable now.

hsoft:
Once again, thanks for this heavy work, much appreciated. I've tested the puppet-lamp scenario myself with this lxdock config:

```
name: puppet-lamp
image: ubuntu/trusty
privileged: true

provisioning:
  - type: puppet
    manifests_path: puppet/manifests
    manifest_file: base.pp
    module_path: puppet/modules
```

There are two problems I encountered.

First, the fact that an `apt update` had never ran prevented puppet from being properly installed. I know that it's not strictly related to this PR, but it's the first time it causes us problems. What would you say about adding an `apt update` run in our debian's `install_packages()`?

Second, `puppet apply` fails here with this message:

```
Error: Puppet::Parser::AST::Resource failed with error ArgumentError: Could not find declared class apache at /.lxdock.d/puppet/manifests/lamp.pp:11 on node puppet-lamp-default-1451369.lxd
Wrapped exception:
Could not find declared class apache
Error: Puppet::Parser::AST::Resource failed with error ArgumentError: Could not find declared class apache at /.lxdock.d/puppet/manifests/lamp.pp:11 on node puppet-lamp-default-1451369.lxd
```

Did you get a similar error @lingxiaoyang ?

lingxiaoyang:
@hsoft Thank you for testing! I confirm the first problem (`apt update`) after destroying and rebuilding the container. I think it's necessary to `update` first. I'll add it.

I didn't replicate problem 2. I attach my logs here:
[puppet-lamp-test.txt](https://github.com/lxdock/lxdock/files/1078739/puppet-lamp-test.txt)

I used the same lxdock file as in your comment. After `apt update`, I got:

````
root@puppet-lamp-default-13681098:~# puppet --version
3.4.3
````

Do you have the same version in the container?

hsoft:
Yes, same version. Since I'm the only one getting the error, I'll do the debugging.

lingxiaoyang:
@hsoft  You need `git submodule init` and `git submodule update`.

hsoft:
Oh, silly me, of course! retrying now.

hsoft:
Yup, it works fine now. For the more complex setups, I'll trust that they work without actually trying them.

Code-wise, this looks good to me. I'd merge. @ellmetha you want to take a look first?

ellmetha:
@hsoft it's OK for me. I didn't have time to test the latest changes in this PR but if you think that we're good to go with these changes, then so do I. 😉 

@lingxiaoyang good work btw!

PR #70, opened 2017-05-05
Enhancement: support shell patterns on `inline` option of shell provisioner

lingxiaoyang:
This pull request proposes the support for shell patterns in inline shell provisioning commands.

Changes proposed:
1. For shell provisioner's `inline` command, use `sh -c 'command_line'` instead of `lxc exec` to always let the shell run the command. The documentation is updated.
2. In `Host.run(cmd_args)`, apply `shlex.quote` on each arguments before joining them with spaces. This will prevent unexpected behaviors caused by a simple space-joining.

### DIFF
--- a/docs/provisioners/shell.rst
+++ b/docs/provisioners/shell.rst
@@ -17,16 +17,12 @@ Just append a ``shell`` provisioning operation to your LXDock file as follows:
 
   provisioning:
     - type: shell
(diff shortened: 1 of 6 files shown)


ghost (approved):
Good stuff, works well, nothing to add. merging, Thanks!

Issue #46, opened 2017-03-20
Better user mapping of shared folders


robvdl:
The way lxdock handles user mapping on shares "sort of" works, but inside the container the files of a share are still owned by NOBODY and NOGROUP which isn't very useful. Any files created on the share in the container will also be created with a strange UID on the host machine like 10000:10000 which isn't very nice either.

The way I dealt with this using LXD before finding LXDock is like this:

* find out your user id and group on host (e.g. 1000:1000)
* put this into /etc/subuid and /etc/subgid on the host (e.g. echo 'root:1000:1' | sudo tee -a /etc/subuid /etc/subgid)
* Add a disk to the container, and set raw.idmap (e.g. lxc config set containername raw.idmap 'both 1000 1000'

It seems LXDock is taking a different approach using ACLs (which I did not need when using my method), but the end result is not very useful, it just makes the share writable but it doesn't seem to solve user mapping issue properly.

hsoft:
The shared folder situation is a tricky one. I know that the solution we've come up with isn't ideal, but it makes many typical workflow work.

Your approach seems interesting, but doesn't mapping root to an existing user circumvent isolation mechanisms supplied by unprivileged containers (that's a real question, I'm not sure of the answer)? If that's the case, we might as well run our containers with `privileged: true`. That makes the shared folder problem much less problematic.

robvdl:
Thanks for the explanation, I will have a look at privileged container over the weekend I think.

I guess this could be the reason why building deb packages in our container comes up with some sort of strange permission error with fakeroot when I do the user mapping my way... I need to dig a bit deeper into how this all works I think.

robvdl:
Unfortunately I found another issue with the way lxdock does ACL's how you end up with an extra -x permission on each file (whether that is the cause I don't know)... but that does not seem to work well with Python's shutil.copytree() function which basically fails because that calls shutil.copystat which then causes an exception when os.setxattr() is called.

hsoft:
A low tech approach to work around this problem would be to add an option to disable ACL fiddling on shared folders. Then, it would only be a matter of having the user run `chmod -R o=Xrw` on the host and in the guest, set `umask 000` in `/etc/profile` to make sure that files created on the mount from the guest OS are accessible to the host. We would then have a somewhat usable two-way shared folder.

But otherwise, as I wrote earlier, it's a tricky problem...

hsoft:
[...]

PR #74, opened 2017-05-08
Only set host side ACL for current user if share is not writable

robvdl:
Because most of the time you will be sharing your current project
source folder, which will already be writable anyway.


hsoft:
I'm not sure I understand the idea behind this change. Of course, your shared folder will always be writable by the host user. The point of the ACL routine is to give read/write access to the *guest*.

robvdl:
As of my last comment on issue #46 I will be redoing this PR with a set_host_acl option on the share instead.

robvdl:
Closing this because it is replace by PR #79

PR #75, opened 2017-05-08
Use stricter variable substitution

robvdl:
Only allow ${var} style variables but not $var anymore.

The latter causes issues with hashed password which then need to be escaped and that gets messy.

That also means that escaping of the $ character is no longer needed.


ellmetha:
Thanks @robvdl! :wink:

Issue #68, opened 2017-05-04
Support LXC "profiles"


lingxiaoyang:
When launching an LXC container, it is possible to set up its profiles that can be managed by `lxc profile` command, for example:

```
lxc launch ubuntu-daily:16.04 docker -p default -p docker
```

LXDock should consider supporting this functionality as it's already [here](http://pylxd.readthedocs.io/en/latest/profiles.html) in pylxd. The YAML definition and the management should be further discussed.

battlemidget:
I think this one can be closed?

hsoft:
Hum, yes, I think so. I haven't followed the development of this feature much, but it looks merged...

PR #77, opened 2017-05-09
New: basic support for LXC profiles. Refs #68

lingxiaoyang:
This pull request allows to set `profiles` in LXDock files, to use existing system-wide profiles.

In the future, LXDock may want to consider project-wise definition, etc.

### DIFF
--- a/docs/conf.rst
+++ b/docs/conf.rst
@@ -193,6 +193,25 @@ Here is an example on how to set up a privileged container in your LXDock file:
 
   Please refer to :doc:`glossary`  for more details on these notions.
 
+profiles
+----------
+
+The ``profiles`` option defines the configuration profiles to use when creating containers. It should
+contain a list of profile names that exist in ``lxc profile list``. Please use ``lxc profile`` command
+to manage profiles as they are system-wide.
+
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+  profiles:
+    - default
+    - docker
+
+  containers:
+    - name: test01
+    - name: test02
+
 protocol
 --------
 
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -12,6 +12,7 @@ _top_level_and_containers_common_options = {
     'lxc_config': {Extra: str},
     'mode': In(['local', 'pull', ]),
     'privileged': bool,
+    'profiles': [str, ],
     'protocol': In(['lxd', 'simplestreams', ]),
     'provisioning': [],  # will be set dynamically using provisioner classes...
     'server': Url(),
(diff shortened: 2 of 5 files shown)


docs/conf.rst
```diff
@@ -193,6 +193,25 @@ Here is an example on how to set up a privileged container in your LXDock file:
...
+The ``profiles`` option defines the configuration profiles to use when creating containers. It should
+contain a list of profile names that exist in ``lxc profile list``. Please use ``lxc profile`` command
+to manage profiles as they are system-wide.
+
+.. code-block:: yaml
+
+  name: myproject
+  image: ubuntu/xenial
+  profiles:
+    - profile1
```
ellmetha: I think it would be better to show a real example here. For example:

```
profiles:
  - default
  - docker
```

lxdock/conf/schema.py
```diff
@@ -12,6 +12,7 @@
     'lxc_config': {Extra: str},
     'mode': In(['local', 'pull', ]),
     'privileged': bool,
+    'profiles': [str, ],
```
ellmetha: I think that profile names have to be real LXD identifiers. Could you use `'profiles': [LXDIdentifier(), ],` instead?

lingxiaoyang: @ellmetha  I'll check it.

lingxiaoyang: @ellmetha  Here's more information on profile name requirements:

General: profile names cannot contain slashes, and cannot be `.` or `..`

Code: https://github.com/lxc/lxd/blob/74e96210ad59b722a602a26579a8b8b561e8251f/lxd/profiles.go#L72-L78
Discussion: https://lists.linuxcontainers.org/pipermail/lxc-devel/2016-August/014576.html

Tests:
```console
$ lxc profile create "a/fa/b/a 2qijgq290 q2398gei@$2132"
error: Profile names may not contain slashes
$ lxc profile create "2qij.gq290 q2398gei@$2132"
Profile 2qij.gq290 q2398gei@132 created
$ lxc profile create "2qij..gq290 q2398gei@$2132"
Profile 2qij..gq290 q2398gei@132 created
$ lxc profile create ".."
error: Invalid profile name '..'
$ lxc profile create "."
error: Invalid profile name '.'
$ lxc profile create "..."
Profile ... created
```

However, I didn't find this in any formal documentation. Perhaps we still leave it as `str`, and let lxd throw the error?

ellmetha: They cannot contain spaces too. :wink: 

```
$ lxc profile create "foo bar"
$ lxc profile list
default
docker
foo
```

This is why I would use the `LXDIdentifier`, which is restrictive enough to ensure consistent profile names (but perhaps too restrictive). Otherwise you could add a `ProfileName` validator.

ellmetha: Oh. It seems you can use profiles name with spaces after all:

```
$ lxc profile create "foo bar"
$ lxc profile list
default
docker
foo
$ lxc delete foo
error: not found
$ lxd delete "foo bar"
Profile foo bar deleted
```

:pensive:

lingxiaoyang: @ellmetha That's in fact a lxd bug...

https://github.com/lxc/lxd/issues/3304

PR #78, opened 2017-05-10
Fixed #71: guest names and ssh setup in Ansible provisioner

### LINKED ISSUE #71: [Bug] Ansible provisioner: wrong guest names and ssh not autostarted for certain guests
In [provisioners/ansible.py](https://github.com/lxdock/lxdock/blob/master/lxdock/provisioners/ansible.py):

1. The names of ArchLinux and OracleLinux should be changed to `arch` (L19) and `ol` (L25) instead of `archlinux` and `oracle`, respectively. The names are used for detecting the distribution ([see the code](https://github.com/lxdock/lxdock/blob/master/lxdock/guests/base.py#L87-L91)) and thus will stay as `arch` and `ol`.

2. On Arch, CentOS, Fedora, openSUSE and Oracle, it seemed that sshd was not started automatically. So we need the similar setup methods for them.

What I tried (needs to be verified further):

```python
    def setup_guest_arch(self):
        # On archlinux guests we have to ensure that sshd is started!
        # https://bbs.archlinux.org/viewtopic.php?id=205367
        self.guest.run(['systemctl', 'enable', 'sshd.socket', '--now'])

    def setup_guest_centos(self):
        # On centos guests we have to ensure that sshd is started!
        self.guest.run(['systemctl', 'enable', 'sshd.socket', '--now'])

    def setup_guest_fedora(self):
        # On fedora guests we have to ensure that sshd is started!
        self.guest.run(['systemctl', 'enable', 'sshd.socket', '--now'])

    def setup_guest_opensuse(self):
        # On opensuse guests we have to ensure that sshd is started!
        self.guest.run(['systemctl', 'enable', 'sshd'])

    def setup_guest_ol(self):
        # On oracle linux guests we have to ensure that sshd is started!
        self.guest.run(['systemctl', 'enable', 'sshd.socket', '--now'])
```

lingxiaoyang:
This pull request fixes guest names, package name, and commands ensuring sshd in Ansible provisioner. 

The only missing guest is openSUSE, which needs too much security configuration that is out of the scope of LXDock, in my opinion.

You may test with the following gist using `lxdock -v up`:
https://gist.github.com/lingxiaoyang/4ae30eed845fe0cccf444d82371abbff

### DIFF
--- a/lxdock/provisioners/ansible.py
+++ b/lxdock/provisioners/ansible.py
@@ -16,13 +16,13 @@ class AnsibleProvisioner(Provisioner):
     name = 'ansible'
 
     guest_required_packages_alpine = ['openssh', 'python', ]
-    guest_required_packages_archlinux = ['openssh', 'python2', ]
+    guest_required_packages_arch = ['openssh', 'python', ]
     guest_required_packages_centos = ['openssh-server', 'python', ]
     guest_required_packages_debian = ['apt-utils', 'aptitude', 'openssh-server', 'python', ]
     guest_required_packages_fedora = ['openssh-server', 'python2', ]
     guest_required_packages_gentoo = ['net-misc/openssh', 'dev-lang/python', ]
     guest_required_packages_opensuse = ['openSSH', 'python3-base', ]
-    guest_required_packages_oracle = ['openssh-server', 'python', ]
+    guest_required_packages_ol = ['openssh-server', 'python', ]
     guest_required_packages_ubuntu = ['apt-utils', 'aptitude', 'openssh-server', 'python', ]
 
     schema = {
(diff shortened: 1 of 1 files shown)


hsoft:
@ellmetha Is the codecov thing really useful to us? For me, it's just noise.

ellmetha:
@hsoft I think it remains useful in order to browse code coverage online. That said pull request comments are a bit noisy IMO. I'll add a ``codecov.yml`` in order to try disabling them.

lxdock/provisioners/ansible.py
```diff
@@ -16,13 +16,13 @@ class AnsibleProvisioner(Provisioner):
     name = 'ansible'
 
     guest_required_packages_alpine = ['openssh', 'python', ]
-    guest_required_packages_archlinux = ['openssh', 'python2', ]
+    guest_required_packages_arch = ['openssh', 'python', ]
```
hsoft: Under arch, `python` refers to python 3. Ansible requires (unless you're bleeding edge and want to use the newly added py3 support in ansible) python2.

hsoft:
@lingxiaoyang to be sure to understand the problem you're fixing: under the 4 distros you're adding code for, sshd wasn't running without that?

lxdock/provisioners/ansible.py
lingxiaoyang: @hsoft I didn't realize the Python versions support... python2 was installed as `python2` binary so I got an error of "Cannot find python". Perhaps I need a symlink, or if there's an Ansible configuration on which Python to use, I can also do it.

lingxiaoyang:
@hsoft No, sshd was not running and I got a connection error when executing `ansible-playbook`.

lxdock/provisioners/ansible.py
hsoft: Oh, that's true. On arch, python 2 has the `python2` executable name. That's a problem. We might as well leave thing as they are and let Ansible run under Python 3 on Arch guests. It *should* work most of the time...

ghost (approved):

PR #79, opened 2017-05-14
Add option to disable setting host ACL on a share

robvdl:
This option defaults to True if not set so the existing behaviour stays the same.

Note: this replaces older PR #74 which can now be closed.

Not sure if a test is needed for this, as it changes very little

### DIFF
--- a/docs/conf.rst
+++ b/docs/conf.rst
@@ -269,8 +269,8 @@ of shared items. Each shared item should define a ``source`` (a path on your hos
 
 .. note::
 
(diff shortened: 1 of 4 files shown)


robvdl:
@ellmetha I agree but should this not be in this section instead, which goes more in-depth on shared folders (just after the section "the problem with shared folder permissions):

http://lxdock.readthedocs.io/en/latest/usage/shared_folders.html

ellmetha:
@robvdl Yes you can also put details regarding this option in the page you mentioned. 😉

ellmetha:
Thanks @robvdl!

PR #84, opened 2017-06-01
Added support for provisioning-specific arguments to the up command

ellmetha:
This PR adds support for using the `up` command as follows:

    $ lxdock up --provision
    $ lxdock up --no-provision

### DIFF
--- a/docs/cli/up.rst
+++ b/docs/cli/up.rst
@@ -1,7 +1,7 @@
 lxdock up
 =========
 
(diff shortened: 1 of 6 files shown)


hsoft:
Looks good to me, but it's a shame to have a broken CI like this. What about temporarily change our `pylxd` requirement to point to the commit corresponding to https://github.com/lxc/pylxd/pull/234 ?

PR #86, opened 2017-06-04
Release v0.3.0

hsoft:
@ellmetha what do you think, is it time yet?

### DIFF
--- a/docs/release_notes/v0.3.rst
+++ b/docs/release_notes/v0.3.rst
@@ -1,6 +1,6 @@
-############################################
-LXDock 0.3 release notes (UNDER DEVELOPMENT)
-############################################
+#####################################
+LXDock 0.3 release notes (2017-07-10)
+#####################################
 
 Requirements and compatibility
 ------------------------------
@@ -23,6 +23,8 @@ New features
   (`#77 <https://github.com/lxdock/lxdock/pull/77>`_)
 * Add a ``set_host_acl`` option to disable setting host ACL on a share
   (`#79 <https://github.com/lxdock/lxdock/pull/79>`_)
+* Add support for provisioning-specific arguments to the ``up`` command
+  (`#84 <https://github.com/lxdock/lxdock/pull/84>`_)
 * Add experimental support for Puppet provisioning
   (`#69 <https://github.com/lxdock/lxdock/pull/69>`_)
 * Add global provisioning capabilities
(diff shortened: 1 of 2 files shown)


ellmetha:
@hsoft Yes, it's time. That said I think that we could wait a few days in order to see if a new version of pylxd is gonna to be released next week. 😉 This would've the benefit of avoiding to use the trick with the special dependency link (for pylxd) in a major release of LXDock.

hsoft:
I was thinking about that too, but the fact of the matter is that lxdock is broken *right now* with LXD 2.14+. I'm not comfortable with that custom dependency link either, but we can easily release v0.3.1 as soon as pylxd's next release is there.

ellmetha:
I am perfectly comfortable with the fact that LXDock is not yet compatible with the newest version of LXDock. This kind of things happen. 'til the problem is solved on the `pylxd` side, I'd say that the v0.3 release can wait. Moreover I'm even less comfortable with the idea of releasing a package that'd need the `--process-dependency-links` to be properly installed.

hsoft:
@ellmetha I'm uneasy about bumping lxc/pylxd#234 again but I'm also uneasy about depending entirely on the resources Canonical has to devote to the project. Here's an idea: What about creating a `pylxd-for-lxdock` package on PyPI? It would be a temporary package until pylxd sees some love again.

robvdl:
Are we mostly waiting for that PR on pylxd to get merged? It seems stuff does get merged but only every now and then.  I am going to ping the author and see where that goes.

robvdl:
I pinged one of the authors who merges commits on pylxd, I didn't want to ping them all because that might seem rude.  I asked nicely, lets's hope they can get this commit through.

ellmetha:
@hsoft it'd be a last resort solution. I'd wait a few days to see if Canonical shows more activity regarding `pylxd`, otherwise let's go with your proposal.

hsoft:
pylxd 2.2.4 is finally released! I'll update this PR.

hsoft:
Here we go, PR is rebased, release date is updated.

hsoft:
@ellmetha I don't know if you're planning on releasing v0.3 right now, but if you do, wait 30 minutes, I have really nice coming that I'd like to slip in v0.3

ellmetha:
@hsoft I'll let you push the release to Pypi then. 😉

hsoft:
Ok, good, because I need another 30 minutes to fix tests.

Issue #88, opened 2017-06-15
Ansible provisioner "inventory" option


jchampion:
Il would be nice to have an option in Ansible provisioner that would allow us to force the inventory path. The inventory that is generated right now is very simplistic and does not deal with groups. I guess you'd have to use only "all" hosts right now for it to work.

While generating the groups based on container names would be possible, the most simple solution is to allow providing the path to an inventory.

Example:
```yml
provisioning:
  - type: ansible
    playbook: deploy.yml
    inventory: inventory/dev
```

hsoft:
@jchampion thanks for your PR! While I very much agree that the current situation is limiting, I'm not sure it's a good idea to go for this solution as a temporary workaround measure.

The problem here is that once we've added a new config parameter, we can't easily remove it without breaking old `lxdock.yml` config, so we'll be stuck with it for a while. If the only reason for its existence is a workaround to temporary limitations (I can't see other uses right now), then we'll give ourselves needless hassle.

Adding grouping to our dynamically generated ansible inventory file doesn't seem overwhelmingly difficult to add. We could add a new `groups` attribute to container elements, and then provisioners can use (or not) this information as they please.

What do you think?

jchampion:
I would not see it as a temporary workaround but more like an advanced option.

Right now, ansible-playbook is executed with a generated inventory that looks like this:
```
X.X.X.X ansible_user=root
```
It does not contain groups, nor does it allow any kind of tweaking such as changing the SSH user/port or any of other ansible_* options (http://docs.ansible.com/ansible/intro_inventory.html#list-of-behavioral-inventory-parameters). That is why I think an "advanced" inventory option would not hurt.


As a sidenote was using only one container until now and I just realized that the provisioning is executed on each container.

Let's say I have an lxdock.yml file such as this one:
```yml
containers:
  - name: app
    hostnames:
      - app.loc
  - name: db
    hostnames:
      - db.loc
provisioning:
  - type: ansible
    playbook: provision.yml
```
ansible-playbook will be executed twice with the "simple" inventory mentioned above. Thing is, it is unlikely that "app" and "db" should be provisioned the same way...

I imagine one of these (or both) could be done:

Option 1
-----------
Generate an inventory that looks like this:
```
[app]
X.X.X.X ansible_user=root
```
ansible-playbook would be executed twice but the group would allow for a specific provisioning.

Option 2
-----------
Execute ansible-playbook only once with a generated inventory that looks like this:
```
[app]
[...]

PR #90, opened 2017-06-22
Add global provisioning capabilities

hsoft:
`provisioning` configuration items declared at the root are now applied
"globally" for provisioners that support it (only `ansible` for now).

Implementation of this feature required quite a refactoring to move the
provisioning logic around. In a nutshell: Less provisioning logic in
`Container` and more in `Provisioner` and `Project`.

One tricky matter was "when can we consider that a specific container
can have its `is_provisioned` flag set?". See code for answer.

I had to change quite many tests, so I used this opportunity to improve
tooling a bit to reduce code duplication in our tests, hence the
introduction of `FakeContainer`. I've only minimally applied its usage,
I'm sure this concept can be expanded and used elsewhere.

ref #88

### DIFF
--- a/lxdock/guests/base.py
+++ b/lxdock/guests/base.py
@@ -182,10 +182,6 @@ class Guest(with_metaclass(_GuestBase)):
         self.run(['tar', '-xf', guest_tar_path, '-C', str(guest_path)])
         self.run(['rm', '-f', str(guest_tar_path)])
 
-    def is_provisioned(self):
-        """ Returns a boolean indicating if the container is provisioned. """
-        return self.lxd_container.config.get('user.lxdock.provisioned') == 'true'
-
     ##################################
     # PRIVATE METHODS AND PROPERTIES #
     ##################################
--- a/lxdock/guests/base.py
+++ b/lxdock/guests/base.py
@@ -76,8 +76,9 @@ class Guest(with_metaclass(_GuestBase)):
     # The `name` of a guest is a required attribute and should always be set on `Guest` subclasses.
     name = None
 
-    def __init__(self, lxd_container):
-        self.lxd_container = lxd_container
+    def __init__(self, container):
+        self.container = container
+        self.lxd_container = container._container
 
     @classmethod
     def detect(cls, lxd_container):
@@ -103,6 +104,12 @@ class Guest(with_metaclass(_GuestBase)):
             if found:
                 break
         return found
+
+    @classmethod
+    def get(cls, container):
+        """ Returns the `Guest` instance associated with the considered container. """
+        class_ = next((k for k in cls.guests if k.detect(container._container)), Guest)
+        return class_(container)
 
     def add_ssh_pubkey_to_root_authorized_keys(self, pubkey):
         """ Add a given SSH public key to the root user's authorized keys. """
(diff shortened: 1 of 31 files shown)


lxdock/guests/base.py
```diff
@@ -175,6 +182,10 @@ def copy_directory(self, host_path, guest_path):
         self.run(['tar', '-xf', guest_tar_path, '-C', str(guest_path)])
         self.run(['rm', '-f', str(guest_tar_path)])
 
+    def is_provisioned(self):
+        """ Returns a boolean indicating if the container is provisioned. """
+        return self.lxd_container.config.get('user.lxdock.provisioned') == 'true'
+
```
lingxiaoyang: @hsoft I think this is mixing up `Guest` and `Container` classes? I see the usage of `Guest.is_provisioned()` method in Provisioner classes, but conceptually this only concerns LXD and is independent from containers' operating systems. 

How about: not passing the provisioned containers' guests into the Provisioner class, inside `Container.provision()` and `Project.provision()`?

hsoft: The proper conceptual place for `is_provisioned()` is fuzzy. `is_provisioned()` is tightly bound to `setup()` because the flag is there partly because we want to ensure that `setup()` is only called once.

The main reason why `is_provisioned()` moved in `Guest` is that it happened *before* I changed `lxd_container` to `container` in the initializer. So at the time, the move was necessary. But then, now that it's not necessary any more, maybe it would be a good idea to move it back in `Container`, yes.

As for your "how about", it's already what we do with the `guests = [g for g in guests if not g.is_provisioned()]`. I don't understand your question.

lingxiaoyang: @hsoft Forget my question. I understood it wrong. I was thinking the Provisioner class completely jumped over the provisioning for provisioned containers, but in fact it only jumps over the setup. This is definitely good to do.

hsoft:
I don't know why there's this flaky setup failure during testing. It seems related to Sphinx, which isn't needed for testing. I'll try to reduce the "flakiness" surface by separating "requirements-dev" and "requirements-tests".

lxdock/provisioners/shell.py
```diff
@@ -15,25 +15,29 @@ class ShellProvisioner(Provisioner):
...
             # Second case: the script is executed on the host side.
             self.host.run([self.homedir_expanded_path(self.options['script']), ])
         elif 'inline' in self.options:
             # Final case: we run a command directly inside the container or outside.
-            host_or_guest = getattr(self, self._side)
+            host_or_guest = self.host if self._side == 'host' else guest
             host_or_guest.run(['sh', '-c', self.options['inline']])
 
+    def setup(self):
+        # nothing to set up, avoid spurious messages with this override.
```
ellmetha: Could this be avoided? Overriding the `setup` method this way in order to shadow setup messages doesn't seem like the right way to do. It shouldn't be required to override the `setup`method this way if you define a provisioner that doesn't involve setup operations IMHO.

hsoft: Three options I see:

1. Removing the logging message at the beginning of `setup_single()`, which allows us to remove the empty override without side effect.
2. Add a `has_setup` class member to `Provisioner` and set it to true in puppet and ansible.
3. Simply remove the override and accept that our shell provisoner will spew a useless "Preparing {0} up for provisioner {1}" message. That's the only effect the removal will have.

All in all, the `has_setup` thing wouldn't be so much more elegant than the current situation, considering that the shell provisioner is likely the only provisioner we'd have that requires no setup.

ellmetha: Yes. Let's wait 'til this situation happen again for another provisioner (which is unlikely anyway).

PR #91, opened 2017-06-22
Make codecov relax

hsoft:
These spurious failures caused by codecov are mightily annoying, even
with comments off. Red X in github are supposed to mean "DO NOT MERGE".
But the thing is, codecov checks are so tight that they trigger alarms
for perfectly good PR, which end up making us *ignore* those big red X
and thus beat the whole purpose behind them.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,9 @@
+coverage:
+  status:
+    project:
+      target: 90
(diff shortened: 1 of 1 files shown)


hsoft:
Examples of PRs that were merged ignoreing the big red X: #78, #79.

PR #94, opened 2017-07-10
Make ansible use the "lxd" transport instead of ssh

hsoft:
The "lxd" transport in ansible allows us to provision our containers
without SSH, directly through "lxc exec" calls!

I had the idea for quite a while that I should explore the possibility
of plugging "lxc exec" into ansible, but I didn't know that the plug was
already there, ready to be used! Had I known that, I would have used it
from the start, it would have saved me SSH fiddling...

Also, you know that small "HOME=" line in tox.ini? I've spent **hours**
debugging this. My tests would fail **only** when running into tox, not
when running directly through pytest. That was very puzzling. Live and
learn...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,11 @@ matrix:
   allow_failures:
     - python: "3.7-dev"
 
+# We set HOME because lxc needs somewhere writable to not crash during testing.
 script:
-  - make travis
+  - cp -R /home/travis/.ssh $TRAVIS_BUILD_DIR
+  - HOME=$TRAVIS_BUILD_DIR make travis
+
 after_success:
   - codecov
 branches:
--- a/docs/provisioners/ansible.rst
+++ b/docs/provisioners/ansible.rst
@@ -84,4 +84,14 @@ This option is a hash of lists of container names. Example:
         group2:
           - container2
     
+lxd_transport
+=============
+
+If this boolean option is set to ``true``, we will use ansible's ``lxd`` transport instead of the
+``ssh`` one, thus bypassing ssh entirely and using ``lxc exec`` directly.
+
+It should be noted that while very cool-sounding, this transport method comes with a couple of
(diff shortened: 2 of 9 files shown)


hsoft:
Travis fails for reasons that seem similar to the problems I had with tox. This will require some debugging. I'll forget the idea of slipping this PR into v0.3.

hsoft:
Oh, I've noticed that ansible actions like `synchonize` don't work with a `lxd` transport. It's inconvenient. I'll have to make this feature optional rather than flat-out replace ssh.

robvdl:
Hi @hsoft, I noticed in a recent change the setup.py was changed to this:

packages=['lxdock']

instead of find_packages() as it was before...

Wow as I recall having done this myself a number of years ago, using just ['lxdock'] won't include subpackages so this will likely break the installer, it also breaks if people build a debian package from this for example it won't include the subpackages lxdock,cli, lxdock.conf, lxdock.guest, etc. in the package which is probably not what you want.

hsoft:
@robvdl I can't seem to reproduce the issue: installing the `master` branch of lxdock in a fresh venv with `pip install .` installs all lxdock's subpackages.

hsoft:
This last commit transforms makes the `lxd` transport optional rather than having it replace the `ssh` transport entirely.

@ellmetha you haven't commented yet. Are you available for reviews? Should I ask @lingxiaoyang instead?

hsoft:
Oh, well, I'm having troubles with re-using the persistent container for SSH-related things. I think I'll have to revert to using fresh containers...

ellmetha:
Hmm the Travis build is still failing though. 🤔

hsoft:
@ellmetha yes, I know. It's puzzling me because it doesn't fail locally on tox. Working on it...

hsoft:
Oh, I think I have it! It seems like it was simply a matter of copying the .ssh folder in `TRAVIS_BUILD_DIR`. I'll try with the "persistent_container" test back on.

PR #95, opened 2017-07-12
Ansible provisioner: add the "groups" option

hsoft:
This allows us to add groups to the dynamically generated inventory.

ref #88


hsoft:
ping @jchampion

PR #97, opened 2017-07-16
Replace ubuntu/xenial with alpine/3.6 in integration tests

hsoft:
Alpine images are 2MB and xenial ones are 90MB!

On my machine, it makes the full test suite go from 356s to 182s, which
I think makes it very worthwhile!


hsoft:
What the heck is this codecov crap? How can this change possibly have decreased code coverage? These metrics are broken, can we disable them?

Issue #109, opened 2017-12-15
Ongoing project maintenance


nkabir:
I've merged pull request #108. @robvdl will have more time in about a month to help. In the meantime, are there any LXDock-specific guidelines for publishing to PyPI and readthedocs.io? Are there any other items to manage for the LXDock development life cycle?

hsoft:
To my knowledge, there isn't much to do other than manage incoming issues and PRs and develop new features to your liking.

robvdl:
We'll figure the release process out in time I think, we need to have a good look at the current state first, when wast the last time there as a pypi release, is the project in a good state for a pypi release or do we still need to fix a number of things first, etc.

I will be available in a week or so... not a month, what I was saying is I have holidays in a weeks time, that last a month so plenty of time coming up.  The last week of work is a bit hectic for me because I work with systems that get really busy during Christmas.

robvdl:
I also wanted to know what to do about the README list of authors (maintainers), it lists the two original authors and then a link to the contributors.  I think this comes up in the docs too and several other places.

The question is what to put there now, I can put my name in that list, do we want the original author names to remain there is upto them.  Should this be a list of active maintainers or past maintainers as well.

hsoft:
sure, no problem. Whatever seems best to you.

robvdl:
I was thinking maybe 2 headings, active maintainers and past or previous maintainers mostly so people know who to contact for the project, as well as have a link to the contributors in github as we have now.

robvdl:
I might also be able to convince maybe two more devs to handle code reviews for me, maybe one person from work as well as a friend who is starting to use lxdock too as a developer.

robvdl:
@hsoft I can't seem to add @johncurley to the group because I am member of the group but not owner.

I want him to help out with my code review queue.

ellmetha:
@robvdl I've just invited him to the group. 😉 

Regarding the list of authors, feel free to change it to whatever seems right to you. But I think that a mention regarding the original authors of this project should be kept somewhere.

robvdl:
Off course I would never remove the original authors, which is pretty much what I said :)

shuhaowu:
Are there still on going maintenance to this project?

robvdl:
[...]

Issue #106, opened 2017-12-08
I don't use lxdock anymore


hsoft:
@ellmetha are you still using it?

I've had second thoughts about the usefulness of creating lxdock. Sure LXD isn't supported on vagrant (there's a new `vagrant-lxd` plugin but it doesn't look like it works well), but I've been working directly with LXC and `vagrant-lxc` recently and, as a matter of fact, it works rather well.

Sure, the whole "box management" part is tricky with LXC/LXD and a `vagrant-lxc` user has to create her own boxes. Sure, LXC is hard to set up properly. But as long as vagrant has the `virtualbox` fallback driver, that doesn't matter much: users will go through complicated setups if they need the speed.

Also, I don't like the idea of using the LXD layer when I don't need it. Also, building Go stuff is a PITA on Gentoo.

I think I'd rather spend my time on `vagrant-lxc` (it doesn't support unprivileged containers yet).

Since I don't see activity on your side either, does that mean that lxdock became unmaintained? If so, we should indicate it in the README. Someone might want to pick it up.

ellmetha:
I don't use it anymore neither. What you said regarding vagrant-lxc makes sense so yes, let's write somewhere that LXDock became unmaintained. 😉

robvdl:
I still use it because I never had much luck with vagrant-lxc in the past and due to it's lack of maintenance and the difficulty I had to install it in the past, is how I initially actually found lxdock in the first place :) I am a bit busy right now with a bunch of Wagtail sites though, but for some crazy reason I always had a burning desire to rewrite it in Go, or is that just madness?  You probably want someone that will continue the existing Python codebase though.

hsoft:
@robvdl if you get to feature parity with the current python version, I don't see the harm in rewriting it (other than what some people would see as needless efforts, but I understand that burning desire: what the hell, let's make useless things!).

But you might want to start a new project altogether. This way, you wouldn't be tied to implementing some feature you might like less about lxdock.

robvdl:
I think lxdock should stay as it is for the time being, keep it Python, if I was to do a rewrite, it will be next year and it will probably be best under another project name entirely.  And yes I probably would only start with implementing the features I cared about the most, which would be to keep it as simple as possible for development containers, I would particularly focus on the dev mount permission mapping because that is what I always struggled with and since I see lxdock as a development only tool, to me it's more important that things work out of the box than being ultra secure, I never liked the whole setfacl thing touching every file in my git repo for example.

[...]

PR #110, opened 2018-02-17
Allow wait timeout to be changed for IP

shuhaowu:
Some containers may be slow to boot up. The default 10 second wait might not be enough.

### DIFF
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -33,6 +33,9 @@ def get_schema():
             'home': str,
             'password': str,
         }],
+        'extras': {
+            'network_wait_timeout': int
+        }
     }
 
     def _check_provisioner_config(config):
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -347,8 +347,9 @@ class Container:
         """ Setup the IP address of the considered container. """
         ip = get_ip(self._container)
         if not ip:
-            logger.info('No IP yet, waiting for at most 10 seconds...')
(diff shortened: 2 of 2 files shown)


hsoft:
Thanks @shuhaowu for your PRs. As we can see in #109 and #106, @nkabir and @robvdl maintain the project now but the project hasn't seen activity in a while. If you don't get a timely review, would you be interested in stepping up as a maintainer?

shuhaowu:
I'm not sure if I should step as a maintainer for now.. as (1) I'm still just evaluating this project for use, (2) I have only a limited amount of time, and (3) I'm not very familiar with the codebase  (or even lxd, for that matter). 

I see some potential use cases for this project outside of just setting up vagrant like development environment (example would be setting up a "reproducible" build environments) and nothing else even come close to what I want. So if I do end up adopting this software, at that point I think I can be better suited to properly "maintain" this.

In the meanwhile, we don't really have to merge any of these PRs if no one is maintaining this. I'll be merging these into my own fork's master regardless.

robvdl:
I like it, but do you think using the key "options" would be better than "extras"?

shuhaowu:
The whole file is about options right? Having options inside options seems strangely unintuitive. `extras` or `advanced` would be somewhat more descriptive as these are extra/advanced tweaks

robvdl:
Ok fair enough. I can merge this now.  Will need to add it to the docs too but I plan to go through the docs before 0.5 release anyway.

PR #112, opened 2018-02-17
Install lxd with snap on travis

### LINKED ISSUE #103: support for snap lxd
Workaround:

```
LXD_DIR=/var/snap/lxd/common/lxd lxdock up
```

Would be nice to pass in an option in `.lxdock.yml` to specify either `LXD_DIR` or that you are running a snap version of lxd since eventually this will be the preferred way of installing lxd.

shuhaowu:
Ubuntu deprecated the PPA and directed people to either use the backports or to use snap. Trusty-backport's version is too old. Thus we try to install this via snap, which means this PR requires #113.

Would be interesting/faster to use the docker version of travis to do this, but I'm not quite sure how it would be possible.

Also closes #103

### DIFF
--- a/lxdock/client.py
+++ b/lxdock/client.py
@@ -1,6 +1,17 @@
+import os
+from pkg_resources import parse_version as ver
+
 import pylxd
 
 
 def get_client():
     """ Returns a PyLXD client to be used to orchestrate containers. """
+    # Fixed a bug upstream: https://github.com/lxc/pylxd/issues/257
+    lxd_dir_not_set = "LXD_DIR" not in os.environ
+    snap_socket_exists = os.path.exists('/var/snap/lxd/common/lxd/unix.socket')
+    insufficient_version = ver(pylxd.__version__) < ver("2.2.5")
+
+    if lxd_dir_not_set and snap_socket_exists and insufficient_version:
+      os.environ["LXD_DIR"] = "/var/snap/lxd/common/lxd"
+
     return pylxd.Client()
--- a/lxdock/client.py
+++ b/lxdock/client.py
@@ -1,7 +1,7 @@
 import os
-from pkg_resources import parse_version as ver
 
 import pylxd
+from pkg_resources import parse_version as ver
 
 
 def get_client():
@@ -12,6 +12,6 @@ def get_client():
     insufficient_version = ver(pylxd.__version__) < ver("2.2.5")
 
     if lxd_dir_not_set and snap_socket_exists and insufficient_version:
-      os.environ["LXD_DIR"] = "/var/snap/lxd/common/lxd"
+        os.environ["LXD_DIR"] = "/var/snap/lxd/common/lxd"
 
     return pylxd.Client()
--- a/lxdock/client.py
+++ b/lxdock/client.py
@@ -1,17 +1,5 @@
-import os
-
 import pylxd
-from pkg_resources import parse_version as ver
 
 
 def get_client():
-    """ Returns a PyLXD client to be used to orchestrate containers. """
-    # Fixed a bug upstream: https://github.com/lxc/pylxd/issues/257
-    lxd_dir_not_set = "LXD_DIR" not in os.environ
-    snap_socket_exists = os.path.exists('/var/snap/lxd/common/lxd/unix.socket')
-    insufficient_version = ver(pylxd.__version__) < ver("2.2.5")
-
-    if lxd_dir_not_set and snap_socket_exists and insufficient_version:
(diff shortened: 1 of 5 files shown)


robvdl:
It looks like just one failing test.  I am wondering in what way is the backports version of lxd "too old", I am still using the version that came with 16.04 and haven't had any issues.

shuhaowu:
I'm still having trouble with the tests on my own machine. Not sure why, but need to check.

Also there is a real failure with respect to LXD_DIR, need to figure that out still.

robvdl:
Ok cool, let me know if you need some help, I will try to get this PR through over the weekend if you can get it passing and it's about time to remove the "this project is not maintained" notice I think :)

shuhaowu:
Strangely enough the tests now passes on my local machine (that's without the pylxd2.2.5 update as well), but refuses to pass on CI.

robvdl:
OK I see what you mean, let me find out what I can do.

It seems to be a recursion loop while installing pylxd in the typing library (Python 3 type hinting).

I wonder if it's OS related, I wonder how many people still care about running the tests on 14.04, or if it is about time to switch to 16.04

I'll take a copy of your commit into another branch and have a play over the weekend if I get a chance.

robvdl:
I have no issues switching to 16.04 minimum because 14.04 never really shipped with lxd in apt in the first place, I doubt many people will be using 14.04 with lxd and going through the trouble to set it up.

Also Ubuntu 14.04 is going out of support April next year and 18.04 is just around the corner, it's about time to move on from Trusty.

Those people still using 14.04 would then have the option to either upgrade to the next LTS, or stick with an old version of lxdock.

But first we need to establish that is the actual problem, at the moment it's just a feeling I have.

robvdl:
I see, your version is probably better than mine and a bit further ahead.

I see you are sleeping until the socket comes up, I ran into that too.

shuhaowu:
Something is wrong with travis I think. It's been acting up.

If you look at https://travis-ci.org/lxdock/lxdock/builds/342664121, the tests ran, but they failed for some reason or another.

robvdl:
Yeah I have been seeing all sorts of weird things with Travis too, for example "apt get remove lxd" and it says it can't do that because the apt cache is locked, but in another python version if will work fine.

Or it fails to pull down the python3.4 tar and says it's corrupt.

Not just that but I've seen all sorts of weird errors yesterday as I was trying things, that seemed unrelated.

shuhaowu:
Not sure if travis actually has proper xenial support tbh: https://github.com/travis-ci/travis-ci/issues/5821

robvdl:
Yeah that is pretty bad, I understand what they are coming from as they have to support a lot of jobs, but it's been 2 years! 2 years since the release of Ubuntu 16.04, and 18.04 is just around the corner.

If we can't figure out how to get this past Travis CI, there are always alternatives like Circle CI etc.

robvdl:
@shuhaowu If you look at issue #106 it seems that another project called conjure-up also run tests with LXD, they are still using Trusty (not Xenial) and the Snap version of LXD so I am going to try that out for a bit as it looks promising. I don't really want to double up on work you've already done in your existing PR but I just want to have a try anyway.

lxdock/client.py
```diff
@@ -2,5 +2,4 @@
 
 
 def get_client():
-    """ Returns a PyLXD client to be used to orchestrate containers. """
```
robvdl: I'm just wondering why remove this docstring, it seems a bit unrelated?

shuhaowu:
> @shuhaowu If you look at issue #106 it seems that another project called conjure-up also run tests with LXD, they are still using Trusty (not Xenial) and the Snap version of LXD so I am going to try that out for a bit as it looks promising. I don't really want to double up on work you've already done in your existing PR but I just want to have a try anyway.

I'll take a look at that again. At this point I'm not sure if you're taking the work of getting the tests running or if I do it. I don't mind either way, but we do have 2 branches up right now.

> They also use a Makefile, we already have a Makefile in the project so could maybe put the build and test related stuff in the Makefile instead of adding a scripts folder, it's just a thought anyway, I don't really mind either way.

I personally don't really like having everything in a giant make file and prefer scripts/* as it is a little easier for people to discover the code that way. I personally associate Makefile to only be for builds (and possibly tests) so I would be surprised if the CI setup code is in there, but that's just my opinion.

lxdock/client.py
shuhaowu: Typo, will fix.

shuhaowu: TBH I find this whole file to be somewhat redundant now that we don't need to do something special to get the client, but yeah.

PR #113, opened 2018-02-17
lxdock + snap

### LINKED ISSUE #103: support for snap lxd

shuhaowu:
Upstream has a bug: https://github.com/lxc/pylxd/issues/257

It is fixed in 2.2.5, which is yet to be released. This commit fixes this situation within lxdock.

Closes #103

### DIFF
--- a/lxdock/client.py
+++ b/lxdock/client.py
@@ -1,6 +1,17 @@
+import os
+from pkg_resources import parse_version as ver
+
 import pylxd
 
 
 def get_client():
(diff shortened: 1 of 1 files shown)


shuhaowu:
Turns out this is required for #112. So I rebased that branch on this one.

Issue #116, opened 2018-03-04
lxdock shell with user defined without home defined tries to set HOME=/root


shuhaowu:
From what I can tell, the way we enter a container is something like:

```
lxc exec container-name [--env HOME=/home/set/in/lxdockyml] -- su -m user
```

Is there a reason why we don't just use `su -l user`? According to man:

       -, -l, --login
           Provide an environment similar to what the user would expect had the user logged in directly.
       -m, -p, --preserve-environment
           Preserve the current environment, except for:

           $PATH
               reset according to the /etc/login.defs options ENV_PATH or ENV_SUPATH (see below);

           $IFS
               reset to “<space><tab><newline>”, if it was set.

           If the target user has a restricted shell, this option has no effect (unless su is called by root).

           Note that the default behavior for the environment is the following:

               The $HOME, $SHELL, $USER, $LOGNAME, $PATH, and $IFS environment variables are reset.

               If --login is not used, the environment is copied, except for the variables above.

               If --login is used, the $TERM, $COLORTERM, $DISPLAY, and $XAUTHORITY environment variables are copied if they were set.

               Other environments might be set by PAM modules.

shuhaowu:
For the record, this is what happens if you have just an user entry under shell (no home defined):

```
$ lxdock shell                                                                                                                         
bash: /root/.bashrc: Permission denied
dev@test-lxdock-test-lxdock-23307078:~$ echo $HOME
/root
```

robvdl:
Thank you, not only have you fixed this, but also #66 which was on my todo list.

PR #117, opened 2018-03-04
 Use raw.idmap to map shared folders' permissions

### LINKED ISSUE #114: lxdock share guest created files not mapped to the host user?
I have a guest created file via the a normal guest user account (non-root). Back on the host, the file has the attributes:

```
-rw-r--r--+ 1 166537 166537  497 Mar  2 11:18 handler.py
```

This means I can't edit this file easily as vim will complain about readonly. How does vagrant-lxc get around this issue?

shuhaowu:
Fixes #114 (lxdock shell doesn't set the correct env vars, commit 1) and #116 (shared folder issues, commit 2).

Fix for #116 is derived from https://stgraber.org/2017/06/15/custom-user-mappings-in-lxd-containers/.

This is kind of a WIP as there's no tests. Some finer points needs to be debated as well. A todo list is as follows:

- [x] We need to have a way to deprecate options in lxdock.yml without breaking old files.
- [x] How do we set /etc/subgid, /etc/subuid? Do we set it from lxdock or do we document this and let the user do it? Right now an error will show up telling user what to do.
- [ ] We need to remove the old documentation and document this new thing.
- [x] Tests.. Not sure if I'll have time to write it, especially considering that I can't consistently get the tests to pass. Perhaps we're better off merging this first and figure out the tests later.
- [ ] How do we clean up those setfacl calls from before? I think `setfacl -Rbn .` will work but this is not researched fully. Not sure if we need to address this, tho.
- [x] Which user do we map to? Right now it maps to either root or the first user in the users config. Not sure if this is a good idea long term, tho.

About setting up your system for this, you need to run:

```
printf "lxd:$(id -u):1\nroot:$(id -u):1\n" | sudo tee -a /etc/subuid
printf "lxd:$(id -g):1\nroot:$(id -g):1\n" | sudo tee -a /etc/subgid
sudo systemctl restart lxd
```

Demonstration of how this works:

```
[thinkpadw530:~/projects/test-lxdock]$ ls -l
total 16
-rw-rw-r-- 1 shuhao shuhao    7 Mar  4 00:25 abc
-rw-rw-r-- 1 shuhao shuhao    5 Mar  4 00:25 bde
drwxrwxr-x 8 shuhao shuhao 4096 Mar  3 23:03 lxdock
-rw-r--r-- 1 shuhao shuhao  192 Mar  4 00:38 lxdock.yml
[thinkpadw530:~/projects/test-lxdock]$ cat lxdock.yml
name: test-lxdock
image: ubuntu/xenial

containers:
  - name: test-lxdock
    users:
      - name: dev

    shell:
      user: dev

    shares:
      - source: .
        dest: /home/dev/share
[thinkpadw530:~/projects/test-lxdock]$ lxdock up
Bringing container "test-lxdock" up
==> test-lxdock: Starting container "test-lxdock"...
==> test-lxdock: No IP yet, waiting for at most 10 seconds...
==> test-lxdock: Container "test-lxdock" is up! IP: 10.28.173.212
==> test-lxdock: Ensuring users are created...
==> test-lxdock: Setting up shares...
==> test-lxdock: share uid map (raw.idmap) updated, container must be restarted to take effect
==> test-lxdock: No IP yet, waiting for at most 10 seconds...
[thinkpadw530:~/projects/test-lxdock]$ lxdock shell
dev@test-lxdock-test-lxdock-23307078:~$ cd share/ && ls -l && cat abc
total 16
-rw-rw-r-- 1 dev dev    7 Mar  4 05:25 abc
-rw-rw-r-- 1 dev dev    5 Mar  4 05:25 bde
drwxrwxr-x 8 dev dev 4096 Mar  4 04:03 lxdock
-rw-r--r-- 1 dev dev  192 Mar  4 05:38 lxdock.yml
adfadf
dev@test-lxdock-test-lxdock-23307078:~/share$ echo "guest" > guestfile
dev@test-lxdock-test-lxdock-23307078:~/share$ ls -l guestfile 
-rw-rw-r-- 1 dev dev 6 Mar  4 05:40 guestfile
dev@test-lxdock-test-lxdock-23307078:~/share$ logout
[thinkpadw530:~/projects/test-lxdock]$ ls -l
total 20
-rw-rw-r-- 1 shuhao shuhao    7 Mar  4 00:25 abc
-rw-rw-r-- 1 shuhao shuhao    5 Mar  4 00:25 bde
-rw-rw-r-- 1 shuhao shuhao    6 Mar  4 00:40 guestfile
drwxrwxr-x 8 shuhao shuhao 4096 Mar  3 23:03 lxdock
-rw-r--r-- 1 shuhao shuhao  192 Mar  4 00:38 lxdock.yml
```

r: @robvdl

### DIFF
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -21,7 +21,7 @@ def get_schema():
             # The existence of the source directory will be checked!
             'source': IsDir(),
             'dest': str,
-            'set_host_acl': bool, # TODO: need a way to deprecate this
+            'set_host_acl': bool,  # TODO: need a way to deprecate this
         }],
         'shell': {
             'user': str,
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -21,7 +21,7 @@ def get_schema():
             # The existence of the source directory will be checked!
             'source': IsDir(),
             'dest': str,
-            'set_host_acl': bool,
+            'set_host_acl': bool, # TODO: need a way to deprecate this
         }],
         'shell': {
             'user': str,
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -25,7 +25,7 @@ def get_schema():
         }],
         'shell': {
             'user': str,
-            'home': str,
+            'home': str,  # TODO: deprecated
         },
         'users': [{
             # Usernames max length is set 32 characters according to useradd's man page.
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -358,12 +358,15 @@ class Container:
 
         logger.info('Setting up shares...')
 
+        if not self._host.has_subuidgid_been_set():
+            raise ContainerOperationFailed()
+
         container = self._container
 
         # First, let's make an inventory of shared sources that were already there.
         existing_shares = {
-            k: d for k, d in container.devices.items() if k.startswith('lxdockshare')}
-        existing_sources = {d['source'] for d in existing_shares.values()}
+            k: d for k, d in container.devices.items() if k.startswith('lxdockshare')
+        }
 
         # Let's get rid of previously set up lxdock shares.
         for k in existing_shares:
@@ -371,25 +374,24 @@ class Container:
 
         for i, share in enumerate(self.options.get('shares', []), start=1):
             source = os.path.join(self.homedir, share['source'])
-            # It is possible to disable setting host side ACL but by default it is always enabled.
-            set_host_acl = share.get('set_host_acl', True)
-            if set_host_acl and source not in existing_sources:
-                logger.info('Setting host-side ACL for {}'.format(source))
-                self._host.give_current_user_access_to_share(source)
-                if not self.is_privileged:
-                    # We are considering a safe container. So give the mapped root user permissions
-                    # to read/write contents in the shared folders too.
-                    self._host.give_mapped_user_access_to_share(self._container, source)
-                    # We also give these permissions to any user that was created with LXDock.
-                    for uconfig in self.options.get('users', []):
-                        username = uconfig.get('name')
-                        self._host.give_mapped_user_access_to_share(
-                            self._container, source,
-                            userpath=uconfig.get('home', '/home/' + username))
-
             shareconf = {'type': 'disk', 'source': source, 'path': share['dest'], }
             container.devices['lxdockshare%s' % i] = shareconf
+
+        guest_username = self.options.get("users", [{"name": "root"}])[0]["name"]
+        host_uid, host_gid = self._host.uidgid()
+        guest_uid, guest_gid = self._guest.uidgid(guest_username)
+        raw_idmap = "uid {} {}\ngid {} {}".format(host_uid, guest_uid, host_gid, guest_gid)
+        raw_idmap_updated = container.config.get("raw.idmap") != raw_idmap
+        if raw_idmap_updated:
+            container.config["raw.idmap"] = raw_idmap
+
         container.save(wait=True)
+
+        if raw_idmap_updated:
+            # the container must be restarted for this to take effect
+            logger.info("share uid map (raw.idmap) updated, container must be restarted to take effect")
+            container.restart(wait=True)
+            self._setup_ip()
 
     def _setup_users(self):
         """ Creates users defined in the container's options if applicable. """
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -122,12 +122,9 @@ class Container:
         shellcfg = self.options.get('shell', {})
         shelluser = username or shellcfg.get('user')
         if shelluser:
-            # This part is the result of quite a bit of `su` args trial-and-error.
-            shellhome = shellcfg.get('home') if not username else None
-            homearg = '--env HOME={}'.format(shellhome) if shellhome else ''
-            cmd = 'lxc exec {} {} -- su -m {}'.format(self.lxd_name, homearg, shelluser)
+            cmd = 'lxc exec {} -- su -l {}'.format(self.lxd_name, shelluser)
         else:
-            cmd = 'lxc exec {} -- su -m root'.format(self.lxd_name)
+            cmd = 'lxc exec {} -- su -l root'.format(self.lxd_name)
 
         if cmd_args:
             # Again, a bit of trial-and-error.
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -389,7 +389,9 @@ class Container:
 
         if raw_idmap_updated:
             # the container must be restarted for this to take effect
-            logger.info("share uid map (raw.idmap) updated, container must be restarted to take effect")
+            logger.info(
+              "share uid map (raw.idmap) updated, container must be restarted to take effect"
+            )
             container.restart(wait=True)
             self._setup_ip()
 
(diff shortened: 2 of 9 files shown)


lxdock/conf/schema.py
```diff
@@ -21,11 +21,11 @@ def get_schema():
             # The existence of the source directory will be checked!
             'source': IsDir(),
             'dest': str,
-            'set_host_acl': bool,
+            'set_host_acl': bool, # TODO: need a way to deprecate this
```
robvdl: Not really, I think we just remove it.  It was me that requested this feature, I doubt anybody else is using it.

robvdl:
I can answer most of these questions:

About breaking old lxdock.yml files, I think since we are just removing the "set_host_facl" setting which is fine as far as I am concerned.  This setting was only added by me in the last release to work around the issue I had with user mapping and I am happy to remove the set_host_facl and do the user mapping properly.

I will update the docs before the next big release, but I don't think it will be a big deal.

I might end up merging all of your commits in a branch first as I also want to add this "default" user functionality still and update the docs and do a a big 0.5 release if this all goes well.

Having to set /etc/subuid and /etc/subgid yeah it's something that is going to be in the docs and README, if you have a message now if these are not set that is fine.

I can sort out the tests if needed, I know it can be annoying sometimes to get tests passing.

We can include the command to cleanup the setfacl in the docs, your command sounds about right but I can look it up just to be sure.

Which user to map to is something I wanted to solve with the "default" user setting I was discussing the other day.  Basically one of the users is the default and if there is no user with default: yes, it creates one with username and password "lxdock".  So in your example you could put default: yes for the dev user and it won't create the lxdock user and the dev user would become the default user, which could be the user used for user mapping, provisioning, ansible, etc.

Another way might be to just select the first user in the list, but I like the idea of introducing a "default user" more and it would solve a lot of problems.

One thing I was wondering, did you have to user privileged containers or not?  It doesn't look like it from your example.

shuhaowu:
No need for privileged container. That was stressed in that article as well.

About the default user: I think we should do something like: default user to lxdock if none is specified, otherwise default to the first user on that list. This would maintain the maximum flexibility i think (suppose someone wants to replicate some other environment and those environment have a default user with a different username, like the travis CI).

shuhaowu:
Also: please do test this branch to see if it works on your setup. 

> I might end up merging all of your commits in a branch first as I also want to add this "default" user functionality still and update the docs and do a a big 0.5 release if this all goes well.

We might want to branch off to a new develop branch, or branch off the current master to a legacy branch as we ramp up the continued maintenance of this project under new people.

robvdl:
I'll just start a new develop branch, the only other branch I can see there is 11 months old from the previous developers so probably no longer relevant considering it's age.

I will test things on my setup don't worry about that :)

robvdl:
The reason the tests have stopped working on Travis is because they were relying on pulling down a PPA version of LXD which has now been deprecated.

It seems this version of LXD is the only version I have ever seen that has "lxc network create ...", which appears to be missing on any other LXD installation I have tried.

I tried going to 16.04 first and used the lxd package from apt, but not much luck with this because it seems that "sudo lxc network create" does not exist, "lxc network" is not a sub-command or so it says.

Sticking with 16.04, I then tried the Snap which gives LXD 2.21 instead of 2.0.11, it seemed to install fine with a bit of messing around and setting the PATH, but still "lxc network create" doesn't seem to exist as a sub-command.

I'm going to take a break from this and try again tomorrow, another way is to get lxd from the backports repo but I have a feeling this will just get the same result so I need to find out what happened to "lxc network create".

I know you can also run the network setup in a console based wizard as well, but that is no good headless with Travis CI.

Here is what I always get, lxc network create command does not exist:

```
$ sudo lxc network create lxdbr0 ipv6.address=none ipv4.address=10.0.3.1/24 ipv4.nat=true
Usage: lxc <command> [options]
This is the LXD command line client.
All of LXD's features can be driven through the various commands below.
For help with any of those, simply call them with --help.
Commands:
  config           Change container or server configuration options
  copy             Copy containers within or in between LXD instances
  delete           Delete containers and snapshots
  exec             Execute commands in containers
  file             Manage files in containers
  image            Manipulate container images
  info             Show container or server information
  launch           Create and start containers from images
  list             List the existing containers
  move             Move containers within or in between LXD instances
  profile          Manage container configuration profiles
  publish          Publish containers as images
  remote           Manage the list of remote LXD servers
  restart          Restart containers
  restore          Restore containers from snapshots
  snapshot         Create container snapshots
  start            Start containers
  stop             Stop containers
  version          Print the version number of this client tool
Options:
  --all            Print less common commands
  --debug          Print debug information
  --verbose        Print verbose information
  --version        Show client version
Environment:
  LXD_CONF         Path to an alternate client configuration directory
  LXD_DIR          Path to an alternate server directory
error: unknown command: network
```

robvdl:
Something interesting I found from a linked issue on the lxd repo for a project called streisand.

They seemed to have gone through the same issue when the PPA got removed and are now using the snap version of lxd, I presume successfully.

They are using Ansible, The .travis.yml file invokes Ansible anf the playbooks gitdy remove the old LXD and put the snap version of LXD on and then start it, which is probably the bit I am missing.

https://github.com/StreisandEffect/streisand/blob/237cc02782d6f2904d645c624ef7009f1e47faa4/tests/development-setup.yml

Hopefully I can get the tests passing again then if there aren't any other obstacles with the newer version of LXD.

shuhaowu:
I assume you're going to get CI working. Do you want me to touch up on the test for this PR so you can merge it after CI works? I should have a little bit of downtime later this week.

shuhaowu:
I updated this branch with the proper unittests. It _should_ all work now. It'll be nice if we can merge this and build on top of it.

robvdl:
I'm about to test this out, sorry it took me a while to get to this.

I might merge this into a dev branch and look at what is needed to fix the tests.

I want this in but I don't think it should be merged into master yet until the tests are passing if possible. I can help with tests as I am pretty keen to get this in myself.

robvdl:
Ah I see it's just coverage but the tests themselves are passing.

robvdl:
I'm going to merge this in master.

I have tested this and this is some awesome work.  I've tested the idmap in the vagrant test box and it's working perfectly.

The next thing I can do if you want, which is to create a default user or lxdock user if no users specified, then set the uid and gid of that user to the host user uid/gid and delete the ubuntu user in the container to avoid possible uid gid clashes.

shuhaowu:
> The next thing I can do if you want, which is to create a default user or lxdock user if no users specified, then set the uid and gid of that user to the host user uid/gid and delete the ubuntu user in the container to avoid possible uid gid clashes.

I'm not sure if i would delete the ubuntu user in the container. The reason is that not all the containers available will necessarily have such a default user.

robvdl:
I will detect if the ubuntu user exists course, the main issue I am thinking about what would happen if 2 users exist with the same uid/gid.

For example if we create a new user called lxdock and make the uid and gid the same as the host OS and the host OS happened to be 1000/1000 (which is pretty common).  Then 2 users in the container will have a uid and gid of 1000.

This is the reason I was thinking why all other users need to be deleted to avoid this, though I am well aware of the multiple OS images that lxdock supports and I was going to test them all.

I might create a new issue to track this work rather than commenting on a closed PR.

shuhaowu:
I don't understand your scenario with multiple users with the same uid. That should be impossible.

robvdl:
I'll start an issue within the next hour and try to explain it there.  I haven't started any work on it yet.

PR #119, opened 2018-03-05
Use Ubuntu 16.04 for Travis instead of 14.04

robvdl:
It seems that the Ubuntu 14.04 LXD repository has dissapeared.

Ubuntu 14.04 is getting quite old now, it's 4 years old and never shipped with LXD in the first place so I am unusure why we use it to run the tests.

So it seems time to set the OS for testing to 16.04 and phase out support for Trusty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,3 @@ __pycache__
 .env/
 env/
 *.retry
-
-vagrant/
-pytest_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ __pycache__
 .env/
 env/
 *.retry
+
+vagrant/
+pytest_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: required
-dist: trusty
+dist: xenial
 before_install:
-  - sudo add-apt-repository -y ppa:ubuntu-lxc/lxd-stable
   - sudo apt-get -qq update
   - sudo apt-get install -y lxd
   - sudo lxd init --auto
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 sudo: required
 dist: xenial
 before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y lxd
+  - snap install lxd
   - sudo lxd init --auto
   - sudo lxc network create lxdbr0 ipv6.address=none ipv4.address=10.0.3.1/24 ipv4.nat=true
   - sudo lxc network attach-profile lxdbr0 default eth0
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 sudo: required
 dist: trusty
-env:
-  - PATH=/snap/bin:$PATH
 
 before_install:
   - make travis-sysdeps
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ sudo: required
 dist: xenial
 before_install:
   - sudo snap install lxd
-  - ls -la /snap
+  - ls -la /snap/bin
   - lxc version
+  - /snap/bin/lxc version
   - sudo lxd init --auto
   - sudo lxc network create lxdbr0 ipv6.address=none ipv4.address=10.0.3.1/24 ipv4.nat=true
   - sudo lxc network attach-profile lxdbr0 default eth0
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,10 @@
 sudo: required
-dist: xenial
+dist: trusty
 env:
   - PATH=/snap/bin:$PATH
 
 before_install:
-  - sudo apt-get remove lxd
-  - sudo apt-get install snapd
-  - sudo snap install lxd
-  - sudo snap start lxd
-  - sudo lxd init --auto --storage-backend dir
-  - sudo lxc network create lxdbr0 ipv6.address=none ipv4.address=10.0.3.1/24 ipv4.nat=true
-  - sudo lxc network attach-profile lxdbr0 default eth0
-  - sudo ls -l /var/snap/lxd/common/lxd
-  - sudo chmod 777 /var/snap/lxd/common/lxd/unix.socket
+  - make travis-sysdeps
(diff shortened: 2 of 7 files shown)


robvdl:
Unfortunately that didn't work, it seems the command that was used to create the network bridge is not present in the Ubuntu 16.04 shipped version of LXD, the version from the PPA that was being used before must have been slightly newer, so I am going to look for another PPA first.

The version from 16.04 seems to be missing the "lxd network create" command, we may be able to work around that, but the easiest solution would be to find another PPA for now.

robvdl:
It seems the official PPA has been deprecated, either switch to the backports repository or the snap.

shuhaowu:
I've already done this work here: https://github.com/lxdock/lxdock/pull/112

robvdl:
Ah thanks, I'll have a look in a sec as I just pushed up another commit that might work, then I'll compare notes :)

robvdl:
I have tried Ubuntu 14.04 with the Snap version of LXD just like the conjure-up project is doing and it's working fine so far, no "unusual" issues with Travis CI like 16.04 was getting so that is great.

I am however running into the strange recursion issue when installing pylxd 2.2.5 (>= 2.2.4 installs 2.2.5 today, but it may not have in the past if 2.2.5 wasn't out yet then, so maybe it still works with 2.2.4)

I will try pinning pylxd to 2.2.4, hopefully it has no issues with this older version.

robvdl:
import recursion loop issue is still happening with pylxd 2.2.4, so going to put it back to >= 2.2.4 as before

robvdl:
I have now solved the recursion issue which actually had nothing to do with pylxd but was caused by the version of sphinx, it could be fixed by pinning sphinx.

Tests are now starting at least, but running into problems, looks like permission errors with lxd.

Once this is solved I want to squash all these commits because 20+ commits of trying things out seems a bit excessive :)

shuhaowu:
> import recursion loop issue is still happening with pylxd 2.2.4, so going to put it back to >= 2.2.4 as before

That version doesn't support snap and requires a workaround that I reverted in my branch.

robvdl:
That's OK I went back to 2.2.5 after I realised the actual recursion loop was caused by the latest Sphinx.  Once I pinned Sphinx it fixed itself.

shuhaowu:
Pinning sphinx seems like the wrong solution to the problem... I would like to simplify some dependencies and just have it install less stuff (like why is this dependent on ipython for development).

robvdl:
Sure there are some things to fix, but right now all I want to see is the tests passing.

I don't know why building the docs is part of travis at all, agreed.

robvdl:
From my last commit I see my problem though, the travis user still isn't part of the lxd group.

Maybe the Makefile thing isn't working well so going to try our script next.

shuhaowu:
You're welcome to try cherry picking my script in the other PR. IF you're not aware already, you can use `newgrp` to login to a group after being added to it.

robvdl:
Haha look at that, the newgrp command seemed to just have stopped the build in it's tracks :) I'm going to have a break for a few hours before continuing but at least I know now what the issue is, it shouldn't be too hard to figure out from here.

robvdl:
There is a huge thread in another project where they try to add the audio group to a user before the tests, this is very similar.

https://github.com/travis-ci/travis-ci/issues/1839

robvdl:
The last error is strange, it seems to be trying to uninstall six, probably to install a newer version and it doesn't like it.

I wonder if an upstream pypi package update has caused this.

Annoying. Now I have to print the groups before that command to see if the sudo -E thing worked.

robvdl:
Yes at least the sudo -E thing worked

    groups
    travis lxd docker

But that pip install failed again, maybe we are not pinning dependencies very well or maybe it is further down the chain but this unrelated pypi package update is now failing the build which is a bit annoying. It should have passed.

robvdl:
I have a feeling my last commit that was pinning py.test is wrong.

Because sudo -E is no longer in the virtualenv which means I am using the wrong python version, the system python.

Earlier in the Travis output I see Travis does this:

    source ~/virtualenv/python3.5/bin/activate

So if we switch to that virtualenv sudo after sudo -E it should do the trick.

shuhaowu:
I feel like using sudo without a virtualenv seems like the wrong approach. Can we just not switch to a different CI provider? I feel like maybe even a VM would be a good way to do this. Perhaps I can even run the tests within lxd itself.

shuhaowu:
We can setup buildkite relatively easily on our own hardware. I maybe able to get some spare capacity to do this.

robvdl:
Possibly, I hadn't heard of it before, but in the mean time I feel like I am getting close now, it's hard to stop.

The sudo -E thing is something I found other projects use to apply to lxd group without the tests stalling which newgrp seems to do.

It may not be the only way though, I haven't looked at other sources.

robvdl:
And it passed.

robvdl:
yay

shuhaowu:
You can keep trying. I'm just looking at alternatives for fun as well. If at any point you feel Travis is a dead end we can possibly switch.


If the problem is the group not being applied to the Travis user and you cannot call newgrp, maybe you can do something silly like `sudo -u travis ....`?

robvdl:
In the mean time it passed which is better than nothing. I need to clean this up and push up another branch as a single commit, it's something at least and it allows me to look at getting the rest of your commits through and focus on that.

I could keep trying variations but I read on that big long thread that if you use sudo -u travis you also have to mention each group as an argument, I didn't look into that though.

shuhaowu:
Awesome work! Thanks a lot for getting this fixed.

robvdl:
I actually forgot the bit that runs the tests is still commented out, I was only starting a container.

Pushing up a final commit.

robvdl:
Now Travis is playing up not starting the build. I might have to try later or in another PR, I think Travis are having some issues right now.

robvdl:
@shuhaowu I have squashed all of these commits in PR #122 and merged into master so that we have a baseline now that everything is working again.

After that I can try putting through cleanup commits to make additional improvements and see if they work, but I think since it was working this needed to be merged first.

We can close this PR now

PR #120, opened 2018-03-08
Working local test setup

shuhaowu:
Since we don't actually have a reproducible working test environment at this point and travis is being extra difficult, I've created a test setup within vagrant-virtualbox (and possibly any arbitrary CI environment, with a common script) such that we can actually run and pass the tests.

It would be nice if we can merge this first so other efforts can continue as we can validate that the existing tests pass. Additionally, having such a local environment is nice as I struggled to pass the tests on my host machine as is.

r: @robvdl


robvdl:
Thanks, I pulled down a copy of this branch locally and am playing around with it over the weekend.  I temporarily changed the box image to ubuntu/trusty64 as that is what travis is using.  I did find I had to provision it in 2 steps, the snap won't install until after a reboot of the vagrant vm for a start but that is OK as this isn't permanent.

Also I had to install some -dev header packages from apt and upgrade pip in the virtualenv first and also python3-virtualenv isn't available on trusty yet either (unless it is in backports) so I had to create the virtualenv the old fashioned way, sigh :)

The tests are passing in my trusty vagrant vm locally, they are still failing with a permission error on travis which is a bit frustrating as I haven't been able to figure out why it means lots of commits trying and retrying until I get it right.

shuhaowu:
If it is possible I'd still like this branch to be merged, so we can have a reproducible setup locally to run tests on. As it stands, the test suite makes a lot of assumptions like the availability of ~/.ssh/id_rsa.

robvdl:
Yes, I might see what I can do to consolidate my Makefile work and this script maybe, but that can be done post merge.  I was just going to look up the bento thing, I haven't used these images before that is all.

robvdl:
Perhaps my vagrant version is too old, I am looking into it.

It tells me that the bento box can't be found and I think this is due to my Vagrant version and v1.8.5 is needed at least.

https://discourse.roots.io/t/cant-access-bento-ubuntu-16-04/8101/2

If so, perhaps the docs should mention that Vagrant 1.8.5+ is needed for local test environment.

The use of vagrant to run the tests and the slogan of lxdock it has above the repo is a bit odd I find :) I have always though of "Vagrant is too heavy" to be a bit negative maybe that isn't needed. But anyways my current Vagrant version appears to be the one that comes with Xenial, I am adding the vagrant repo now and trying this again.

shuhaowu:
I can try getting this to run within lxdock, but after #117 merges tho so file share works as expected.

robvdl:
I need to go for a few hours but have tested with the latest vagrant and working fine as expected.

I like this, I might expand a bit on this later but this is fine for now.  I found the first thing I did after vagrant up is to run make install then make spec.

But there is also a tox way of running the tests.  Either way, it might be useful to run make install as part of provisioning.

robvdl:
I'll merge this but just a few things I found, it would be nice that the provisioning runs make install which is needed in to run the tests, I can add that in myself.

Also I am getting just a few tests fail, not sure why, I tried to reboot the VM and run the tests but still the same thing.  Most tests are passing except a few.

robvdl:
@shuhaowu RE the "some tests are failing" comment I made earlier, I manage to fix that but I also got Vagrantfile and travis to share the LXD setup code in the Makefile so I don't need ci-base-setup.sh anymore.  I will push something up soon, I have an existing PR I can just add another commit to that.  It just avoids having similar code in 2 places and now both travis and vagrant are sharing the same LXD setup code.

shuhaowu:
I'm still not a big fan of having the setup script within a Makefile. It just feels like the wrong place to be placed into. Plus, it makes things like if/while statements super awkward to write.

robvdl:
Yeah I was thinking about that, I mostly used the Make because the Makefile was already there from the previous devs but I am not sure how they worked and ran tests locally.  For some reason your script, after the VM was provisioned some tests were failing for me, not sure why but we can compare and find out.

PR #121, opened 2018-03-10
Tests are working again on Travis using snap

robvdl:
This commit is a squashed version of last 40 commits in PR #119

* Use snap version of lxd and continue to use Trusty for now as it seems more reliable
* Fix import ordering in test_project.py that seems to have come from an updated version of isort
* Move Travis setup to Makefile where lxd is installed from snap and a network setup
* Needed to use sudo -E workaround to apply lxd group to travis user before running tests
* Pin version of Sphinx used because latest version was going into a recursion loop during pip install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,6 @@
-sudo: required
-dist: trusty
-before_install:
-  - sudo add-apt-repository -y ppa:ubuntu-lxc/lxd-stable
-  - sudo apt-get -qq update
-  - sudo apt-get install -y lxd
-  - sudo lxd init --auto
-  - sudo lxc network create lxdbr0 ipv6.address=none ipv4.address=10.0.3.1/24 ipv4.nat=true
-  - sudo lxc network attach-profile lxdbr0 default eth0
-  - sudo chmod 777 /var/lib/lxd/unix.socket
-  - ssh-keygen -t rsa -b 2048 -f ~/.ssh/id_rsa -P ""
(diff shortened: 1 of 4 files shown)


robvdl:
See PR #119

I need to merge this into master now because the Travis build has been broken for far too long.

There our outstanding PR's that can be merged if the tests are passing again.

.travis.yml
```diff
@@ -22,13 +12,28 @@ matrix:
...
+  - sudo addgroup lxd || true
+  - sudo usermod -a -G lxd $USER || true
+  - sudo ln -s /snap/bin/lxc /usr/bin/lxc
+
 # We set HOME because lxc needs somewhere writable to not crash during testing.
+# The "sudo -E" line is needed because the user was added to the lxd group which needs to be applied.
 script:
   - cp -R /home/travis/.ssh $TRAVIS_BUILD_DIR
-  - HOME=$TRAVIS_BUILD_DIR make travis
+  - sudo -E sudo -u $USER -E bash -c 'source ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/activate; HOME=$TRAVIS_BUILD_DIR make travis'
```
shuhaowu: This feels like an awful hack, heh.

robvdl: Yeah it does, I can try alternative ways in separate PRs to fix that later, now that the tests are running to see if I can improve that.

PR #124, opened 2018-03-11
Local Vagrant test setup and Travis using the same ci setup scripts

robvdl:
A number of small commits to cleanup and consolidate Travis CI setup and Vagrant local test setup.


robvdl:
@shuhaowu if we use your script instead of the Makefile, I need to remove the install python3-virtualenv from it. Because Trusty doesn't have python3-virtualenv, only Xenail does and I'd like to run the script on both. That is why I moved the install python3-virtualenv line into the Vagrantfile itself.

Additionally I need to work out why not all the tests were passing when the Vagrant box was provisioned with the script, but they are all passing with the Makefile.

robvdl:
I'll have a look what I can do to go back to that script instead of Makefile so long I can make the tests pass.

Another thing I was thinking about (but won't do that in this PR), is multiple Python versions and Tox.

The Travis environment runs against multiple Python versions, I am not even sure if we need that for the local tests but we could potentially add deadsnakes and install python 3.4 and 3.6 as well so that tox can be run in the Vagrant VM.

robvdl:
That concludes this commit, I am ready to merge this.

I have restored your script ci-base-setup.sh rather than use the Makefile so that it should suit both our needs hopefully.

Tests are working both on Travis and locally.

robvdl:
I'm going to squash and merge this now because I want to build on top of this and it gets tests working locally.

Since I have restored the ci-base-setup.sh script and removed lxd setup code from the Makefile this should be fine.

PR #125, opened 2018-03-11
Avoid installing Sphinx during Travis tests (fixes recursion issue)

robvdl:
It passed in the Vagrant VM (16.04) but I need to know if it works on Trusty on Travis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
-.PHONY: install spec upgrade lint coverage isort travis docs
+.PHONY: install install-tests spec upgrade lint coverage isort travis docs
 
 install:
 	pip install -r requirements-dev.txt
+	pip install -e .
+
+install-tests:
+	pip install -r requirements-tests.txt
 	pip install -e .
 
 upgrade:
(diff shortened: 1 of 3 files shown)


robvdl:
As I suspected, what happened here is that Sphinx 1.7 was released and that seems to be causing recursion issues during pip install but only on Trusty.  This does not happen on the Vagrant Xenial test VM.  Either we pin Sphinx < 1.7 or if it is possible rework things to avoid installing sphinx during the Travis tests.

robvdl:
I have moved flake8 and isort to test requirements.

By doing this Travis is not needing to install Sphinx anymore, which isn't needed to run the tests.

robvdl:
Basically this commit avoids installing Sphinx on Travis because we don't need Sphinx in order to run the tests and newer versions of Sphinx aren't working on Trusty.

Sphinx was pinned to an older version in an earlier commit because it seems to have issues installing Sphinx 1.7 on Trusty and Travis CI going into an infinite recursion loop.

This PR removes the version pin again, putting it back as it was, and moves a few dependencies to the test requirements file so that Sphinx doesn't get installed on Travis.

Sphinx will still be installed on the local vagrant test VM, just not on Travis.

PR #126, opened 2018-03-11
Tidyup .travis.yml

robvdl:
Don't create symlink in /usr/bin

I don't think this is needed anymore, it should all be handled in ci-base-setup.sh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,6 @@ env:
   global:
     - PATH=/snap/bin:$PATH
 
-before_script:
-  - sudo addgroup lxd || true
-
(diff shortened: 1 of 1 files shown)


robvdl:
I did this in a few commits to make sure that Travis didn't need any of these lines.

Will squash into a single commit on merge.

robvdl:
Yep that is correct, thanks for review.

robvdl:
Actually not quite correct, you mean "no need to create the lxd group manually because that is done by the snap package". and "user is already added to the lxd group in scripts/ci-base-setup.sh"

PR #134, opened 2018-03-15
logger.warn is deprecated, use logger.warning

robvdl:


robvdl:
@johncurley this is a pretty small commit, if you could look over it I'll merge it, ignoring the coverage drop failing the build, it's just a bit too sensitive sometimes.

robvdl:
As you can see, Travis is passing the commit but codecov is a bit too sensitive.

PR #135, opened 2018-03-15
Fix shell -c command is now like Vagrant

### LINKED ISSUE #129: lxdock shell -c "command" in quotes produces strange error
Without quotes:

    $ lxdock shell -c ls /
    bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  src  srv  sys  tmp  usr  va

With quotes:

    $ lxdock shell -c "ls /"
    /.lxdock.d/shell_cmd.sh: 2: /.lxdock.d/shell_cmd.sh: ls /: not found

robvdl:
The shell command feature did not seem to work when running two commands because it was using shlex.quote on each argument.

See: issue #129

With the current syntax I was unable to work out how to simply run two commands in a container, Vagrant uses this much simpler syntax which makes much more sense:

    vagrant ssh -c "command1; command2"

Also Vagrant will produce an error if you add another argument so this will not work:

    vagrant ssh -c command1 command2

It seemed lxdock followed something like that except it auto quoted each argument, it was quite confusing and I couldn't figure out how two run two commands and that produced an error.

I have changed lxdock to be the same as how vagrant works:

    lxdock shell -c "command1; command2"

Fixed tests and updated zsh config, the bash one doesn't need updating by the looks of it.

Closes #129

### DIFF
--- a/contrib/completion/zsh/_lxdock
+++ b/contrib/completion/zsh/_lxdock
@@ -94,10 +94,10 @@ case $state in
         ;;
 
       (shell)
-        # lxdock shell name [-h] [-u USERNAME] [-c ...]
+        # lxdock shell name [-h] [-u USERNAME] [-c COMMAND]
         _arguments '::container:__container_list' \
                    '(-u --username)'{-u,--username}'[Username to login as]:username:' \
-                   '(-c --command)'{-c,--command}'[Command to be executed]:*command:command:'
+                   '(-c --command)'{-c,--command}'[Command to be executed]:command:'
         ;;
 
       (status)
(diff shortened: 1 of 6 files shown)


robvdl:
Note if you are wondering why in some tests I am also removing "home" from shell, that is explained in this commit, shell home is deprecated:

cd8c032d61e5ad00f597a222dec52523056c9ec5

robvdl:
@nkabir @johncurley @shuhaowu If you could please look over this commit as there are a few tests that needed fixing.

shuhaowu:
:+1: I agree this is a much more obvious interface than the previous.

nkabir:
Looks good. Agreed with @shuhaowu that it improves interface clarity.

PR #137, opened 2018-03-18
Updating docs for 0.5 release part 1

robvdl:
* Mostly updating the get_started.rst page and documenting Snap installation method
* Also add recommendation about installing lxdock in a virtualenv to discourage sudo pip installs
* Add note about migrating from apt to Snap version of LXD without going into too much detail
* Add section to getting_started.rst about setting up the host for shares and remove setfacl reference
* Mention Vagrant test box in contributing.rst page
* Don't be too assertive about coverage (exclamation mark), yes we want to be strict, but sometimes a small drop is unavoidable and should be discussable on an issue
* Found one typo in glossary.rst

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -22,14 +22,14 @@ Prerequisite: install LXD
 You may want to skip this section if you already have a working installation
 of LXD on your system.
 
-For Debian or Ubuntu 16.04 and higher if you prefer to stick with apt:
+LXD is available in the repository for Debian or Ubuntu 16.04 and higher:
 
 .. code-block:: console
 
   $ sudo apt-get install lxd
 
 You can now also install LXD from Snap which works on Ubuntu 14.04 and higher.
-Since the PPA has been deprecated, this is now the easiest way to get
+Since the LXD PPA has been deprecated, this is now the easiest way to get
 the latest version of LXD on Ubuntu.
 
 .. note::
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -37,7 +37,7 @@ the latest version of LXD on Ubuntu.
   If you have already installed and configured LXD from apt earlier and
   want to upgrade to the Snap version, you may need to purge LXD packages
   first and reboot for the old network bridge to be removed. You can migrate
-  existing containers using ``lxd.migrate`` after or just start fresh.
+  existing containers using ``lxd.migrate`` or just start fresh.
 
   .. code-block:: console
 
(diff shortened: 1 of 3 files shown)


robvdl:
@johncurley could you please look at part 1 of the docs update for the new 0.5 features we've been adding.

I want to do the second part as another PR but this mostly does the getting started section.

docs/getting_started.rst
```diff
@@ -20,32 +19,46 @@ LXDock should build very easily on Linux provided you have LXD available on your
 Prerequisite: install LXD
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You may want to skip this section if you already have a working installation of LXD on your system.
+You may want to skip this section if you already have a working installation
+of LXD on your system.
 
-For Debian and Ubuntu, the following command will ensure that LXD is installed:
+For Debian or Ubuntu 16.04 and higher if prefer to stick with apt:
```
robvdl: The reason I put 16.04+ here for the apt version, is because 14.04 no longer has access to the PPA so will have to use the Snap on 14.04 now

docs/getting_started.rst
```diff
@@ -20,32 +19,46 @@ LXDock should build very easily on Linux provided you have LXD available on your
...
 
-For Fedora, LXD is available through an experimental COPR repository. Unfortunately SELinux is not
-yet supported, therefore make sure it is disabled or set to permissive. Then run:
+For Fedora, LXD is available through an experimental COPR repository.
+Unfortunately SELinux is not yet supported, therefore make sure it is
+disabled or set to permissive. Then run:
 
 .. code-block:: console
 
-    # dnf copr enable ganto/lxd
```
robvdl: I'm not sure why this was using # and the others $ so I made it consistent.

docs/getting_started.rst
```diff
@@ -86,6 +117,13 @@ You should now be able to install LXDock using:
 
   $ pip3 install lxdock
 
+.. note::
```
robvdl: It would be nice if we could provide Snap packages and maybe debs or flatpak eventually.  It would make the setup instructions a lot simpler, and this manual setup part can be moved down then.

robvdl:
@johncurley I made quite a few commits on this PR due to typos and rewording, so if you could look at the diff of all the commits together and review it, then I can squash merge it.

docs/getting_started.rst
shuhaowu: `#` implies a root shell. `$` implies a non-root shell. The original instruction is correct and the new ones are not.

docs/getting_started.rst
shuhaowu: Also this would avoid the issues of lxdock not being available after activating a virtualenv/conda env.

PR #139, opened 2018-03-29
Add installation method on Fedora using snaps

robvdl:
Thanks @johncurley for adding this and testing


robvdl:
Several things need to be added:

* The Fedora SNAP method seems to work fine with SELinux and doesn't encourage the user to disable it, so it might actually be best to recommend the SNAP method first even on Fedora, rather than last.
* The Fedora COPR method is missing the lxd-client dependency

@johncurley  mentioned he might be able to send a patch to the COPR repository maintainer to sort the missing SELinux rules out.  But it does seem that the SNAP package is better maintained than the COPR method so we really should be mentioning it first instead of last.

PR #143, opened 2018-04-26
Initial x11 container implementation

shuhaowu:
WIP for now.

There're some issues with .Xauthority as I don't quite understand that file and seem like it might not exist?


shuhaowu:
For NVIDIA runtime mounting into the container, see https://github.com/lxc/lxd/commit/65cec7f2c884ea2490fe9376d10abe8aefa97b84

However, the version of LXD that has this feature is only available in lxd 3 for now.

robvdl:
@shuhaowu seems to need a rebase, is this still WIP?

shuhaowu:
Yeah still not sure if this is the right approach for the long term. I am trying a plugin system right now.

PR #145, opened 2018-04-30
Expanduser for shares as well as additional share properties.

shuhaowu:
Allows shares to be specified with ~ on the host.

This allows one to share things like .ssh/.gnupg and other stuff easier into the container.

Additionally, I'm adding an entry to allow the specification of additional disk options as documented by https://github.com/lxc/lxd/blob/master/doc/containers.md#type-disk

@robvdl

### DIFF
--- a/lxdock/conf/config.py
+++ b/lxdock/conf/config.py
@@ -75,7 +75,7 @@ class Config:
         config.interpolate()
 
         try:
-            schema(config._dict)
+            config._dict = schema(config._dict)
         except Invalid as e:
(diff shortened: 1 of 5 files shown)


shuhaowu:
The error here is due to ansible now requiring python 3.5+. Perhaps we should either fix the version of ansible or stop building for 3.4?

shuhaowu:
Any updates on this? I'm working on a number of features for lxdock and this is going make it difficult to maintain my patches to be upstreamable as there are so many of them.

PR #152, opened 2018-06-25
Allow users to become sudoers automatically

shuhaowu:
This PR introduces a `sudoer` option for each user, which will automatically add the user to the list of passwordless sudoers in the container. It is accomplished via files in /etc/sudoers.d, which should be supported in most distros afaik.

This should eliminate some unnecessary provisioning step and streamline the process. 

r: @robvdl

### DIFF
--- a/lxdock/conf/schema.py
+++ b/lxdock/conf/schema.py
@@ -32,6 +32,7 @@ def get_schema():
             'home': str,
             'password': str,
             'shell': str,
+            'sudoer': bool,
         }],
         'extras': {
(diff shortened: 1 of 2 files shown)


robvdl:
I like this but would personally have called the config option "sudo: true" rather than "sudoer: true".

I'm a bit late to the party to merge this though, two years have passed since.

Do you have an objection renaming this config option?

mtausig:
This branch should be rebased to the current master, to check if it passes the tests.

PR #153, opened 2018-06-25
Correctly set the SHELL variable

shuhaowu:
Right now, since we are launching shell commands by first transferring a script and then launching the script as if it is a shell (see man su for the -s option), it sets the SHELL environment variable to the script.

Some applications relies on using the SHELL environment variable to launch subprocesses, which could result in a forkbomb. This PR fixes this behaviour by setting the SHELL to /bin/sh, which is accurate.

r: @robvdl

### DIFF
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -140,6 +140,7 @@ class Container:
             self._container.files.put(self._guest_shell_script_file, textwrap.dedent(
                 """\
                 #!/bin/sh
+                export SHELL=/bin/sh
                 {}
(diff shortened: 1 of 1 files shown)


robvdl:
I tested this patch against the current master and just running into two failing tests.

I might create a copy of this PR so I can play with it, perhaps the tests can be fixed.

robvdl:
@shuhaowu fyi, I am merging this but as PR #173 instead because that one passes.

closing this one

PR #155, opened 2018-06-26
Fix Travis CI tests, lxd init already sets up a bridge now

robvdl:
Tests were failing because the bridge already existed.


robvdl:
Seems to be hanging in the Travis tests now, could be related to the LXD 3 release?

robvdl:
Yeah but it is still failing CI due to a bug in LXD 3.x (I believe) which is making all container operations extremely slow and makes the tests time out when run on Travis. They are still passing in the Vagrant VM at least which is good, but it just takes a very long time for them to complete, too long for Travis.

PR #163, opened 2019-10-13
See shell provisioner stdout and stderr

### LINKED ISSUE #104: I don't get any stdout/stderr from shell provisioning
I am guessing lxdock only captures output on the host side, but ... it would be nice with output from shell provisioning too :)

shuhaowu:
This allows the stdout/stderr of the shell provisioner to be displayed in the terminal in real time. Fixes #104.

Using a similar technique, I can get ansible running on the container itself without having to ssh. I'll be working on this soon (tomorrow?) as I need this functionality for my setup.

### DIFF
--- a/lxdock/guests/base.py
+++ b/lxdock/guests/base.py
@@ -158,12 +158,15 @@ class Guest(with_metaclass(_GuestBase)):
     # HELPER METHODS #
     ##################
 
-    def run(self, cmd_args):
+    def run(self, cmd_args, stdin_payload=None):
         """ Runs the specified command inside the current container. """
         logger.debug('Running {0}'.format(' '.join(cmd_args)))
-        exit_code, stdout, stderr = self.lxd_container.execute(cmd_args)
-        logger.debug(stdout)
-        logger.debug(stderr)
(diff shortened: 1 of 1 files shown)


robvdl:
OK we might just have to merge this as I doubt CI will pass, but the PR itself looks fine.

    ./tests/unit/provisioners/test_ansible.py:32:18: W605 invalid escape sequence '\s'
    make: *** [lint] Error 1

I've seen this before in other projects, newer versions of flake8 not liking an escape character in a regext that usually needs to be made a raw string.

robvdl:
On that note, I've been thinking of ditching Python 3.4 because it's long out of support and lots of Python libraries are dropping it recently, hardly anything works on Python 3.4 and Trusty anymore going from experience.

robvdl:
We can easily fix the issue with the invalid escape sequence \s but I would do it as a separate PR, as there could be additional issues stopping CI from running as it hasn't been run for a while now.

shuhaowu:
https://github.com/PyCQA/pycodestyle/issues/766

Will be fixed along with the ansible provisioner.

Issue #161, opened 2019-02-08
pylxd 2.2.9 breaks lxdock


lingxiaoyang:
The change in pylxd 2.2.9 regarding `container.execute()` seems to break the command execution of lxdock and it is breaking the initialization. 

Here's what I captured in logs of `lxdock -v up` with pylxd 2.2.9:

```
Global provisioning with ansible
Preparing default up for provisioner ansible
Installing packages required for the ansible provisioner on the guest
Running apt-get update
http://localhost:None "POST /1.0/containers/modele-default-29654/exec HTTP/1.1" 202 997
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 934
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 934
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 934
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 934
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 934
http://localhost:None "GET /1.0/operations/b6c428ec-0feb-478e-a281-abfde2caaad7 HTTP/1.1" 200 464
Get:1 http://security.ubuntu.com/ubuntu xenial-security InRelease [109 kB]


Running apt-get install -y apt-utils aptitude python
http://localhost:None "POST /1.0/containers/modele-default-29654/exec HTTP/1.1" 202 1035
http://localhost:None "GET /1.0/operations/dcdef0cc-35b8-465f-afe8-c468c6a5cc3f HTTP/1.1" 200 972
http://localhost:None "GET /1.0/operations/dcdef0cc-35b8-465f-afe8-c468c6a5cc3f HTTP/1.1" 200 972
http://localhost:None "GET /1.0/operations/dcdef0cc-35b8-465f-afe8-c468c6a5cc3f HTTP/1.1" 200 972
http://localhost:None "GET /1.0/operations/dcdef0cc-35b8-465f-afe8-c468c6a5cc3f HTTP/1.1" 200 972
[...]

PR #166, opened 2020-01-17
Allow different types of SSH keys

### LINKED ISSUE #165: SSH keys with other algorithms should be supported
Currently, SSH key deployment into the container uses a hardcoded key name (`~/.ssh/id_rsa.pub`). Thus keys using another algorithm (ed25519, ecdsa) are not detected.

mtausig:
This should fix #165.
Unfortunately, I haven't been able to properly test it on my system due to issue #161 .

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@
 # requires Vagrant 1.8.5+ for the bento image. Run "make coverage" inside box.
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "bento/ubuntu-16.04"
+  config.vm.box = "bento/ubuntu-18.04"
 
   config.vm.provision "shell", path: "scripts/ci-base-setup.sh", privileged: false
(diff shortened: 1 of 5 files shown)


robvdl:
what about id_dsa.pub

mtausig:
Since DSA is very much deprecated, I decided against including it.
But if you prefer, I can add it.

robvdl:
I see, I'm not sure then. We can leave it out and if anyone asks we just mention dsa keys are deprecated.

mtausig:
I can add a comment.
BTW: Is there something like a recommended dev environment for new contributors to get the current master branch actually running. I'd like to test this myself.

robvdl:
Well we created a simple Vagrantfile for basic testing of the application a year or so ago, it's nothing too fancy but we used it for testing in the past as I recall.

It's been a long time since I've worked on this project at all though, so can't say if it still works now.

mtausig:
Comment added & tested (at least manually).

This is finished IMO.

PR #168, opened 2020-01-24
Fix travis build

mtausig:
Travis build currently fails.

### linting 

The linting in the travis build failed, due to a newer flake8 version. See https://github.com/MathSci/fecon236/issues/6 for reference.
Prefixing the string holding a regex with `r` (according to the current python conventions) fixes that.

### pytest version

Newer versions of pycov require a newer version of pytest.

### alpine update

The tests used *alpine/3.6* as a test image, which does not exist anymore in the standard remote. Version bump to 3.11.

The new version of alpine requires a root password to be set manually, otherwise SSH (and thus ansible) won't work.

### Use newer pylxd version

In the past, the pylxd version used (in the tests) was restricted to 2.2.7 (due to an upstream bug mentioned in #155), but this broke with the introduction of #163 which used newer API signatures.
The upstream bug has been fixed, so an update to at least 2.2.10 is possible

### Indentation error

One schema test had an indentation error. I guess it never worked before.

### Python 3.4

The deprecated python version 3.4 has been removed from the build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: true
 dist: trusty
 
 python:
-  - "3.4"
   - "3.5"
   - "3.6"
   - "3.7-dev"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,13 @@ dist: bionic
 python:
   - "3.5"
   - "3.6"
-  - "3.7-dev"
+  - "3.7"
+  - "3.8"
 
 matrix:
   allow_failures:
-    - python: "3.7-dev"
+    - python: "3.7"
+    - python: "3.8"
 
 install:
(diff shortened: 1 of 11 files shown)


robvdl:
Looks like just down to one test failing now, good work!

I think Python 3.4 should just be removed too from Travis CI tests as Python 3.4 is out of support since early last year and almost nothing seems to work on it anymore :)

robvdl:
If you can't get that last test passing after having a go at it for some time, I am not opposed to temporarily skipping this one test with a description why. The reason being that a working and passing CI suite is far more important, it helps everyone else who contributes with their pull requests which are all failing now.

mtausig:
I think this one is done. All tests work fine.

Issue #122, opened 2018-03-10
Documentation tasks for 0.5


robvdl:
[ ] Document idmap changes
[ ] Document network timeout (extras)

mtausig:
With #169 merged, we can check the second last box.

robvdl:
Yeah we really should try to make the 0.5 release happen. I got so busy at work over the last year the lxdock project just didn't move (until recently again).

robvdl:
See #184

PR #169, opened 2020-01-26
Fix and document Vagrant usage

mtausig:
I have updated and fixed the usage of the Vagrant box for the test environment and added documentation about its usage, too (which closes part of #122 ).

### DIFF
--- a/scripts/vagrant-base-setup.sh
+++ b/scripts/vagrant-base-setup.sh
@@ -2,13 +2,16 @@
 
 set -xe
 
-# python3-virtualenv is not available on Trusty so install it here
+# Install packages to use a virtualenv with python3
 sudo -E apt-get -y install python3-setuptools python3-dev python3-virtualenv
 
 # create a virtualenv and install lxdock
 python3 -m virtualenv -p python3 ~/venv
 source ~/venv/bin/activate
+
 cd /vagrant
+# Clean local builds to enable fresh build inside the Vagrant box
+make clean
 make install
 
 # automatically activate virtualenv and switch to /vagrant on login
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@
 # requires Vagrant 1.8.5+ for the bento image. Run "make coverage" inside box.
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "bento/ubuntu-16.04"
+  config.vm.box = "bento/ubuntu-18.04"
 
   config.vm.provision "shell", path: "scripts/ci-base-setup.sh", privileged: false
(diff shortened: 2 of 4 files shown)


mtausig:
I will squash & rebase this, once #168 is merged.

robvdl:
Ok I'll wait for your rebase then review this a bit later, thanks.

mtausig:
Rebase done.

scripts/vagrant-base-setup.sh
```diff
@@ -2,13 +2,15 @@
 
 set -xe
 
-# python3-virtualenv is not available on Trusty so install it here
-sudo -E apt-get -y install python3-setuptools python3-dev python3-virtualenv
+sudo -E apt-get -y install python3-setuptools python3-dev python3-virtualenv python3-pip
 
 # create a virtualenv and install lxdock
 python3 -m virtualenv -p python3 ~/venv
+pip3 install pip
```
robvdl: Should this be "pip3 install -U pip"? so that it upgrades to the latest

robvdl: To be honest I don't quite understand why the statement "pip3 install pip" is there as we are creating a virtualenv the line above it (on line 8) and then activating it on line 10. What I would have thought to see is upgrading pip just after activating the virtualenv maybe? But why install pip with pip3 globally I wonder what the purpose is for that?

mtausig: The problem was, that the binary name in Ubuntu for the python3 version of pip is `pip3`, but the Makefile calls just `pip`. So it was either changing the Makefile, creating a symlink to pip3 or installing it in the virtualenv.

robvdl: Ok that is really odd because as far as I know as soon as you activate the virtualenv, pip is available from venv/bin/pip regardless if python3-pip is installed from apt. I can have a play with it and see but from my perspective it seems odd to do any pip install commands before the virtualenv is activated, they should appear after the virtualenv is activated so that we know we are using pip from the virtualenv not the system /usr/bin or /usr/local/bin

robvdl: I'm pretty sure the way it's activating the virtualenv now doesn't work like that in scripts that is probably why if you use "pip" it won't find it because the path isn't set but I need to verify this. From my experience, using source like that in a script doesn't work

mtausig: You are right. I just tested it on a fresh machine, it is really not necessary. I will remove it.

mtausig: 1a3fa4f78396031c9947e373d132a87dc4ba11af

PR #170, opened 2020-02-08
Install SSH key in authorized_keys of all users created by lxdock

### LINKED ISSUE #167: Deploy SSH keys to user accounts
Currently, the SSH key from the host machine is only deployed to the `authorized_keys` file of the root user in the container.
If additional users are created in the container by lxdock, it would be desirable to make it an authorized key for them as well.

mtausig:
Previously, the SSH key from the host user was only installed in the authorized_keys file of the root user in the container. It only did that, if the ansible provisioner was used.

This path changes two thing:

* The key is always installed in the container, even without ansible being used
* The key is installed for all users created by lxdock, not just root

This fixes #167

### DIFF
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -411,17 +411,24 @@ class Container:
     def _setup_users(self):
         """ Creates users defined in the container's options if applicable. """
         users = self.options.get('users', [])
+        ssh_pubkey = self._host.get_ssh_pubkey()
+
         if users:
             logger.info('Ensuring users are created...')
             for user_config in users:
                 config = copy.copy(user_config)
                 name = config.pop('name')
                 self._guest.create_user(name, **config)
+                if ssh_pubkey is not None:
+                    # Use the default home directory for the SSH key, unless explicitely configured
+                    home_dir = '/home/'+name
+                    if 'home' in config:
+                        home_dir = config['home']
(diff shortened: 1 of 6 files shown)


mtausig:
Rebased to master.

robvdl:
Thanks for fixing up the flake8, I know they can be annoying sometimes.

Just isort now and it should pass, I normally have a "make fix-isort" in my own Makefiles but that doesn't seem to be in the lxdock Makefile, you can add it if you want. Basically the same as "make isort" without the --check-only

mtausig:
Done

robvdl:
I've tested it and going to merge now. Just had to install openssh-server in the container which it currently doesn't do by default off course.

I can't remember, but did it used to install openssh-server if the Ansible provider was used before perhaps?

mtausig:
> I can't remember, but did it used to install openssh-server if the Ansible provider was used before perhaps?

Yes. The ansible provisioner automatically installs openssh-server.

I still think it makes more sense to installk the keys independent from ansible, that's why I changed the logic (plus, it was easier to implement that way)

robvdl:
Ah, I might add a PR then that installs openssh-server then.

I don't see the point copying ssh keys to the machine if openssh-server isn't going to be installed :)

mtausig:
Not sure if this is necessary. Most people will install it anyway, and otherwise the keys, while not of any use, at least don't hurt.

PR #173, opened 2020-02-10
Correctly set the SHELL variable

robvdl:
Right now, since we are launching shell commands by first transferring a script and then launching the script as if it is a shell (see man su for the -s option), it sets the SHELL environment variable to the script.

Some applications relies on using the SHELL environment variable to launch subprocesses, which could result in a forkbomb. This PR fixes this behaviour by setting the SHELL to /bin/sh, which is accurate.

Based on original PR#153 by shuhaowu, checking to see if we get two failed tests agains latest master.

Thank you for contributing to LXDock! A list of simple rules is available on the online
documentation to help you contribute to this project: http://lxdock.readthedocs.io/en/latest/contributing.html.
Please review these guidelines before submitting your pull request!

### DIFF
--- a/lxdock/container.py
+++ b/lxdock/container.py
@@ -140,6 +140,7 @@ class Container:
             self._container.files.put(self._guest_shell_script_file, textwrap.dedent(
                 """\
                 #!/bin/sh
+                export SHELL=/bin/sh
                 {}
                 """.format(command)))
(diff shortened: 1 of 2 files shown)


robvdl:
Hi @shuhaowu, if you're OK we can merge this into master.

I just picked up your PR and fixed two tests which were failing.

robvdl:
I think we could just merge this and close the other PR

It's exactly the same + the test

PR #178, opened 2020-02-14
Update uidmap documentation

mtausig:
Update to the documentation how folders are shared. 
Fixes the 4th box of #122 

Questions for @robvdl to complete it:

There are two share related attributes defined in the schema: `set_host_acl` and `share_properties`. I couldn't figure out completely what they do (if anything at all).

`set_host_acl` appears to be deprecared, I couldn't find it in any python file.

`share_properties` is parsed in the `_setup_shares` method of  `lxdock/container.py`, but I think it is just some sort of passthrough to pylxd. What can you do with it?

### DIFF
--- a/docs/conf.rst
+++ b/docs/conf.rst
@@ -275,12 +275,6 @@ of shared items. Each shared item should define a ``source`` (a path on your hos
   shares:
     - source: /path/to/my/workspace/project/
       dest: /myshare
-      set_host_acl: true
-
-.. note::
-
-  The ``set_host_acl`` parameter is optional and defaults to true when left out,
-  please refer to :doc:`usage/shared_folders` for more information.
 
 shell
 -----
--- a/docs/usage/shared_folders.rst
+++ b/docs/usage/shared_folders.rst
@@ -31,19 +31,30 @@ changing them in the host as well, and vice versa. That leaves us with a problem
 solve gracefully. Things become more complicated when our workflow has our container create files in
 that shared folder. What permissions do we give these files?
 
-LXDock tries to answer this by using ACLs. To ensure that files created by the container are
-accessible to you back on the host (and vice versa), every new share has a default ACL giving the
-current user full access to the source folder. An ACL is also added for the root user of the
-container in order to allow him to access the shared folders on the guest side with read/write
-permissions.
+There are two possible ways for making a shared folder actually usable on both ends:
+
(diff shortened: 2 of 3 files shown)


robvdl:
I'll have to think about it as it's been a while, but I was definitely involved around the time we did the host_acl thing. As for share_properties, I'm not sure.

The host_acl setting had to do with the share and idmap thing, can't quite remember if it was deprecated, I'll have to have a look at the history and it should all come back to me.

shuhaowu:
You shouldn't use host_acl anymore if you can do idmap. idmap is much smoother.

mtausig:
`set_host_acl` is completely unused. I removed it from the schema and the documentation.

mtausig:
> You shouldn't use host_acl anymore if you can do idmap. idmap is much smoother.

I disagree here, idmaps have their drawbacks. But that' not really relevant for this PR, maybe a future discussion.

mtausig:
What do we do with `share_properties`?

* Remove it, since we don't know what it is doing
* Ignore it
* Add a documentation line mentioning it without describing

robvdl:
Note: If the PR is failing only due to codecov then it can still be merged, codecov has always been a bit sensitive and adding lots of docs probably triggers it.

The main issue with the old method (share_properties?) before we used idmap as I recall, is it recursively had to set extended permissions on each file on the share. This interfered with some things in the container, building debian packages was one of those things I had found. But something else also didn't like it, forgot what that was now sorry.

When switching to idmap, a project with existing shares should remove these extended permissions recursively. At the time (over 2 years ago) I had a command for that which should probably be in the docs as an upgrade path step, I'll need to dig it up, perhaps it is in the comment history on an old PR or issue.

mtausig:
> Note: If the PR is failing only due to codecov then it can still be merged, codecov has always been a bit sensitive and adding lots of docs probably triggers it.

"Problem" is there, because I deleted a single line of code (the obsolete item from the schema), which slightly reduced (theoretically) the coverage

mtausig:
I suggest to merge this as is, and add an issue which tracks the unknown state of *share_properties*.

mtausig:
Rebased

mtausig:
Ping

robvdl:
Ah yeah I'll try to fit it in the weekend, things have been a bit crazy in NZ with the country going into lockdown sorry.

mtausig:
Ping

mtausig:
Ping

Issue #146, opened 2018-05-09
UserWarning: Attempted to set unknown attribute "location" on instance of "Container"


tannoiser:
Hi.

I don't know if it's a real issue or just a warning or a problem of mine environment, but I got this error message:

```
mizio@ubuntu:~$ lxdock up
/home/mizio/.local/lib/python3.5/site-packages/pylxd/models/_model.py:116: UserWarning: Attempted to set unknown attribute "location" on instance of "Container"
  key, self.__class__.__name__
Everything is already up and running! Nothing to do.
```

some info on my environment:

```
mizio@ubuntu:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.4 LTS
Release:        16.04
Codename:       xenial

mizio@ubuntu:~$ python3 --version
Python 3.5.2
mizio@ubuntu:~$ lxdock --version
lxdock 0.4.1
mizio@ubuntu:~$ lxc --version
3.0.0
```

lxd is installed from snap. lxdock is installed following your docs (with pip3 by my user).
I can't say if is a problem or not: it quite works as you can see:

```
mizio@ubuntu:~$ lxc list
+------+-------+------+------+------+-----------+
| NAME | STATE | IPV4 | IPV6 | TYPE | SNAPSHOTS |
+------+-------+------+------+------+-----------+
mizio@ubuntu:~$ cat lxdock.yml 
name: mytest
image: ubuntu/xenial

containers:
   - name: myproject01
   - name: myproject02
   - name: myproject03
     image: archlinux
mizio@ubuntu:~$ lxdock up
Unable to find container "myproject01" for directory "/home/mizio"
Creating new container "mytest-myproject01-520193" from image ubuntu/xenial
Unable to find container "myproject02" for directory "/home/mizio"
Creating new container "mytest-myproject02-520193" from image ubuntu/xenial
Unable to find container "myproject03" for directory "/home/mizio"
Creating new container "mytest-myproject03-520193" from image archlinux
Bringing container "myproject01" up
Bringing container "myproject02" up
Bringing container "myproject03" up
==> myproject01: Starting container "myproject01"...
==> myproject01: No IP yet, waiting for at most 10 seconds...
==> myproject01: Container "myproject01" is up! IP: 10.186.210.73
==> myproject02: Starting container "myproject02"...
==> myproject02: No IP yet, waiting for at most 10 seconds...
[...]

PR #179, opened 2020-02-28
Package lxdock as a snap

### LINKED ISSUE #156: turn lxdock into a snap
Feature request to turn lxdock into a snap, I could probably take a stab at this shortly.

mtausig:
Instructions how to build a use a snap package of lxdock.

I also updated the *LXD installation* part of the documentation, to primarily suggest a snap based installation, since that is the official method Cannonical suggests and it also allows a more easy integration on other distributions than Ubuntu.

I would suggest to merge this as a first step as is, and later on adapt the Travis configuration to automatically build and publish an official package.

Fixes #156

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ upgrade:
 	pip install -e . -U
 
 lint:
-	flake8
+	flake8 --exclude snap,parts,stage,prime
 
 isort:
 	isort --check-only --recursive --diff lxdock tests
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: install install-tests spec upgrade lint coverage isort travis docs clean
+.PHONY: install install-tests spec upgrade lint coverage isort travis docs clean snap
 
 install:
 	pip install -r requirements-dev.txt
@@ -28,6 +28,11 @@ travis: install-tests lint isort coverage
 
 docs:
 	cd docs && rm -rf _build && make html
+
+snap:
+	snapcraft clean lxdock -s pull
+	snapcraft
+	snapcraft clean lxdock -s pull
 
 clean:
 	find . -name '*.pyc' -exec rm -r {} \;
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,20 @@
+name: lxdock
+version: git
+summary: lxdock
+description: |
+  Build and orchestrate your development environments with LXD 
+  - a.k.a. Vagrant is Too Heavy
+base: core18
+confinement: strict
+
+parts:
+  lxdock:
+    plugin: python
+    python-version: python3
+    source: .
+    requirements: requirements-dev.txt
+apps:
+  lxdock:
+    command: bin/lxdock
+    plugs:
+      - lxd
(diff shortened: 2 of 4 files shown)


mtausig:
Continuing the discussion with @robvdl from https://github.com/lxdock/lxdock/issues/146#issuecomment-584509575 :
All the dependencies (all pip pacakges) are installed with the snap. So unless we rebuild our packages, they are not going to change.

Makefile
```diff
@@ -13,7 +13,7 @@ upgrade:
 	pip install -e . -U
 
 lint:
-	flake8
+	flake8 --exclude snap,parts,stage,prime
```
robvdl, line 16: can go in setup.cfg the exclude stuff, I thought we already had some there

snapcraft.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - a.k.a. Vagrant is Too Heavy
+base: core18
+confinement: strict
+
+parts:
+  lxdock:
+    plugin: python
+    python-version: python3
+    source: .
+    requirements: requirements-dev.txt
```
robvdl: am I reading this right? do the dev requirements end up in the snap? why? we don't need them in the snap or any other kind of binary distribution like deb/rpm

mtausig:
So the requirements-test.txt would be the appropriate choice?

robvdl:
Neither, the way this project is actually setup:

* setup.py has dependencies needed to run the application "so pip install ." should do it
* requirements-test.txt contains extra dependencies only relevant to run the tests (like tox, pytest)
* requirements-dev contains extra dependencies only needed for developing lxdock (right now that is just sphinx and ipython

so really you just need setup.py in order to run the app

Makefile
mtausig: But `setup.cfg` doesn't get used if I just run `make lint` on the command line during development.

snapcraft.yaml
mtausig: FIxed: 5e261b3

mtausig:
Rebased.
I'll squash this once you give the ACK.

mtausig:
Ping

mtausig:
Ping

robvdl:
I'm in the middle of an import project with fixed deadline, likeliless I can get to it before November is quite slim.

Makefile
robvdl: I'll have a look when I get back in November but yes the flake8 tool does read setup.cfg for config if there is a [flake8] section. If we are to publish snaps using travis it has to be on merge, but to be honest I am not a huge fan of snaps and didn't really want to get involved in maintenance of snaps but I'll have a look in early November. Problem with the project is I'm the only person still merging stuff for the last 2 years and I'm insanely busy sometimes it takes a while as I like to test everything properly.

robvdl:
I'm so sorry for all the work that has gone into this and then it not even getting merged. I tried for a while to look after this project but it just turned out to be impossible by myself. I'm sorry for this and being a let down to not seeing this through and getting merged.

We did get others putting their hand up to help years and years ago but nothing ever came of that, even then it seemed to be just me maintaining it.

I now have not even run or tested lxdock for over 3 years and have created a ticket to move the entire repo to archival status and declare the project dead, just nobody even seems to care about that anymore because a year on and not a single comment on that ticket.

